PR #3, opened 2023-03-06
Correct notebook docs

Winterhuman:
Mentions possible `intel_agp` issues with hibernation, and that the three `ACTION=="add"` rules are not needed for kernels newer than `linux 5.5`.


ptr1337:
Thank you. Did not notice this PR.
I merge it.

PR #4, opened 2023-04-12
Wiki: Update and expand specific pages

Lousio:
- Updated scheduler in kernel page and changed up the wording a bit
- Kernel page updated to latest information
- Add EEDVF in the supported scheduler list
- Add Bottles in Gaming section
- Add Protonup-qt as a manager for Wine and Proton
- Add clarification regarding different runners and when to use which

Feedback always welcome.

### DIFF
--- a/home/Gaming.md
+++ b/home/Gaming.md
@@ -73,6 +73,8 @@ Script Installers
 
 For many games, installation is made easy with script installers.
 
-For example, the [League Of Legends scripts](https://lutris.net/games/league-of-legends/) can be installed with minimal effort. We recommend using the WINE Standard version. Simply follow the instructions provided by the script.
+For example, the [League Of Legends scripts](https://lutris.net/games/league-of-legends/) can be installed with minimal effort. Simply follow the instructions provided by the script. Generally, it is recommended to use Wine-GE for non-Steam games while Proton-GE, Proton-CachyOS and Proton-Experimental are excellent choices for Steam games. Try to not use Proton for non-Steam games as Proton relies on the Steam runtime.
+
+A couple of games require a specific, patched Wine-GE version to run (e.g. League of Legends).
 
 With this guide, you are now ready to start gaming on CachyOS. Enjoy!
--- a/home/Gaming.md
+++ b/home/Gaming.md
@@ -50,8 +50,6 @@ You can use Proton to play your Windows games on Linux.
 
 CachyOS provides various Proton versions for improved performance, including `proton-cachyos`, `proton-ge-custom`, `proton-tkg-git`, and the official Proton versions `proton-experimental` and `proton`.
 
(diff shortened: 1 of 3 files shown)


home/cachyos-performance.md
```diff
@@ -8,11 +8,11 @@ editor: markdown
 dateCreated: 2021-10-04T07:59:10.433Z
 ---
 
-# Linux Kernel with BORE scheduler is the default choice
+# Linux Kernel with EEDVF scheduler as the default scheduler
 
-Yes, this is a part of the optimization for CachyOS. ¿Why do we use that scheduler in particular?. Well... let us explain why!. There are so many schedulers like BMQ, MuQSS, that some of you might already know (or not), but we have tested BORE and it resulted the one with the best performance results both in benchmarks and responsiveness tests.
+An important aspect of optimization for CachyOS is the scheduler used in our default kernel. Let us explain why we have chosen the EEVDF as our default scheduler. There are many schedulers available, like BMQ or MuQSS, some of them you might already know. We have tested EEVDF and it resulted the one with the best performance results both in benchmarks and responsiveness tests. Please keep an eye out for any updates, as schedulers tend to change often and our default scheduler might change in the future.
```
ptr1337, line 13: Can you please move out MuQSS and BMQ?
I don't think this should be in there. MuQSS is obsolete at all since more then a year.

home/Gaming.md
```diff
@@ -50,8 +50,7 @@ You can use Proton to play your Windows games on Linux.
 
 CachyOS provides various Proton versions for improved performance, including `proton-cachyos`, `proton-ge-custom`, `proton-tkg-git`, and the official Proton versions `proton-experimental` and `proton`.
 
-Bottles
--------
+Additionally, [Protonup-qt](https://aur.archlinux.org/packages/protonup-qt) is a easy way to install and manage [Wine-GE](https://github.com/GloriousEggroll/wine-ge-custom) and [Proton-GE](https://github.com/gloriouseggroll/proton-ge-custom) installations depending on game launcher.
```
ptr1337, line 53: We don't recommend to use protonup-qt in performance aspects.
Better point to the packages which are in the repo:
proton, proton-ge-custom, proton-cachyos, proton-experimental. These are compiled against systemlibs and should bring a more smooth experience.
 Same for wine-ge-custom :)

ptr1337 (changes requested):
Please check the comments I have added.
Besides that im good with them! Thanks for your PR.

home/cachyos-performance.md
Lousio: Replaced with BORE and PDS, which are more known.

home/Gaming.md
Lousio: It was mainly added as some people would prefer a GUI alternative, has been taken off with commit 7e2e5ce.

ptr1337:
Now im good with.
Sadly @SoongVilda did commit some stuff. I tried to revert it, but can't merge it still.
Can you rebase it please?

Lousio:
Seems like it can be merged now, it's missing some parts from my text in the gaming one but that can be changed afterwards.

PR #11, opened 2024-03-25
fix: correct directory name for amd pstate status and prefcore

GreyXor:
the amd-pstate module is actually called amd_pstate


ventureoo (approved):
Thanks

PR #12, opened 2024-04-07
Update how_to_add_cachyos_repo.md

HyoAkira:
I edited wget command to curl, also it's not necessary to syu, it should be  pacman sy because some updates may break system.


ptr1337:
Could you maybe make also a sperate PR for our rework wiki?

Ive already added this page.

https://github.com/CachyOS/wiki/tree/wiki-rework

PR #15, opened 2024-04-21
Boot managers page

Pogogo007:


ptr1337 (approved):
LGTM, but I'd maybe move systemd-boot and refind over grub, since people will likely just not read the full document.

PR #19, opened 2024-05-29
Just new style update

ghost:
If needed can also remake content

### DIFF
--- a/src/content/docs/configuration/kernel-manager.md
+++ b/src/content/docs/configuration/kernel-manager.md
@@ -5,13 +5,11 @@ description: The CachyOS Kernel Manager is a powerful tool for managing kernels,
 
 import ImageComponent from '~/components/image-component.astro';
 
-CachyOS Kernel Manager
-======================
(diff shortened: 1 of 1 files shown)


vnepogodin:
That won't work. That file is MDX so it has .mdx ext

ghost:
> That won't work. That file is MDX so it has .mdx ext

I changed it from mdx to .md

ptr1337:
> > That won't work. That file is MDX so it has .mdx ext
> 
> I changed it from mdx to .md

Actually, I think then importing pictures wont work. MDX is required for that.

ghost:
> > > That won't work. That file is MDX so it has .mdx ext
> > 
> > 
> > I changed it from mdx to .md
> 
> Actually, I think then importing pictures wont work. MDX is required for that.

Should i just remove them or what to do with that?

vnepogodin:
> > That won't work. That file is MDX so it has .mdx ext
> 
> I changed it from mdx to .md

yes. we need .mdx there

PR #25, opened 2024-06-13
Rewrite some of the pages

1Naim:
First PR!


src/content/docs/installation/filesystem.md
```diff
@@ -126,5 +130,10 @@ Bcachefs is still considered as experimental and can have issues.
 - Setup can be complicated
 
 ## TL:DR
-Use **xfs** or **ext** by default, BTRFS if snapshots and compression is something you want and ZFS if the ARC cache or zpools are wanted.
+Use the default filesystem **BTRFS** as it is considered stable and has a lot of neat features (snapshots, compression, etc). Use **XFS** or **Ext4** for a simple
+and fast filesystem.
+
+:::note
+Filesystem performance between BTRFS, XFS, Ext4 are negligible.
```
ventureoo, line 137: Well, to be honest, that's not really true. CoW-based systems can have noticeable performance losses compared to a classic FS like Ext4 for a number of reasons, but primarily because of the overall complexity of approach taken.

1Naim: should have probably added "on NVMEs". 

edit: or probably just removed altogether

PR #28, opened 2024-06-20
Begin work on Support: Getting Help.

1Naim:
This PR starts the work on the "Getting Help" page and adds Discord and CachyOS Forums. I do not use Matrix nor Telegram so someone else has to add each of those sections

### DIFF
--- a/src/content/docs/support/getting_help.md
+++ b/src/content/docs/support/getting_help.md
@@ -1,3 +1,28 @@
 ---
-title: todo
+title: Getting Help
 ---
+
+CachyOS has a very friendly and active community that can help you if you encounter an issue. You can
+reach out for help through one of the services we support.
+
+:::note
+We have a webhook that bridges messages from Discord, Matrix and Telegram to each other.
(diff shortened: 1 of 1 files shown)


vnepogodin:
IMHO the difference between some pages(getting help, troubleshooting, submitting bugs) seems not quite clear
https://github.com/CachyOS/wiki/blob/next/src/content/docs/support/submitting_bugs.md

and here is all the links, except matrix
https://wiki.cachyos.org/policy/social

1Naim:
hmmm, should we unify "Troubleshooting" and "Submitting Bugs" while renaming "Getting Help" to something else like "Contact Us" or "Join Our Community" or "Socials"?

I think this makes each section more clear while not separating related stuff, e.g. when you want to submit a bug report the first thing to do is troubleshoot whats up.

edit: Troubleshooting just seems like a very vague concept to put into the wiki since the methodology differs from each problem.

1Naim:
Now that I look at it again, is "Getting Help" even necessary?  I think we can just move https://wiki.cachyos.org/policy/social to Support rather than Policy

PR #30, opened 2024-06-28
pics for the installation process

Froggyman-kashee:
#staykashee


lambdakilo (changes requested):
These are great although a bit outdated already at this point (not your fault). What part of the Wiki uses these btw?

PR #39, opened 2024-07-15
gaming: rework game_performance to better reflect actual functionality

1Naim:
`game-performance` actually changes power profiles, the CPU governor changing is just a byproduct of that. I also added a note for `intel_pstate` because it sets EPP to performance and not the governor itself. This has confused a user in the past


vnepogodin (approved):
LGTM

PR #41, opened 2024-07-18
add grub-btrfs usage to post_install md

heheda123123:


Pogogo007:
There is an option in cachyos hello to automatically create snapper configs. The grub btrfs part is fine however

heheda123123:
nice, I have not found it before. Can add this to wiki.

ptr1337:
For grub-btrfs-support, there is a package which sets this up, called "grub-btrfs-support" please change to that.

PR #51, opened 2024-08-12
feat: Overhaul Bootloader Configurations

1Naim:
This PR moves bootloader configurations to its own page and expands upon each configuration possible, e.g. kernel cmdline config, loader config, bootloader theming.

I also renamed bootmanagers -> bootloaders to keep consistency.


ptr1337 (approved):
Thank you! Looks good to me!

vnepogodin (approved):
legit

PR #53, opened 2024-08-16
docs: Address the Differences Between Boot Loaders and Boot Manager

1Naim:
There are two parts to this PR. The first part is renaming all instances of boot loaders to boot managers. The second is adding a preambule to the boot manager installation page to differentiate between the two terms.

When talking about UEFI systems, the correct term to refer to systemd-boot, rEFInd and even GRUB
is boot manager. A boot manager is not the same as a boot loader.

The motivation of this PR is the big interest I gained after reading through some of Rod Smith's web pages regarding EFI boot loaders, found [here](https://www.rodsbooks.com/efi-bootloaders/). 

Some can argue that this can be considered nitpicking. While I do agree partially, I am also stand by the fact that using the correct terms must be necessary as to not cause misinformation and even confusion.


aarrayy:
LGTM

Aarrayy (approved):
LGTM

ptr1337 (approved):

PR #59, opened 2024-08-23
Complete overhaul of Installation Prepare

Aarrayy:
I felt it was needed in order to be easier to follow and i deleted some non-needed parts too.

Some parts were reused but not many, and the reasoning on why i moved to Getting Started is because they're just preparations before installation and it didn't make much sense to have it on the "installation" index.


ptr1337:
Perfectly Aarrayy! Thank you very much

PR #62, opened 2024-09-04
Sched ext Overhaul

Aarrayy:
This serves as a guide & introduction to the sched-ext schedulers.

bpfland, rusty and lavd were added and considered since they're the most up to date and mature (for now).

I will keep updating this section according to any new change or feature.

Feel free to suggest any change.


ptr1337 (approved):

src/content/docs/configuration/sched-ext.md
```diff
@@ -122,13 +223,19 @@ Options:
...
+
+Mainly because they are (for now) multipurpose schedulers, meaning they adapt to many workloads even if they don't excel at all of them.
+
+In order to find out which one fits you best, there is no other shortcut than to test it yourself.
+
 ## GitHub
 
-- scx-scheds (Schedulers): https://github.com/sched-ext/scx
-- https://github.com/sched-ext/scx-kernel-releases
+- scx-scheds (Schedulers): <https://github.com/sched-ext/scx>
```
1Naim: I don't really understand why ur doing this for links. Can you elaborate?

src/content/docs/configuration/sched-ext.md
```diff
@@ -9,16 +9,18 @@ build another kernel for a different scheduler.
...
 
-##  Installing a Kernel with sched-ext support
+## Installing a Kernel with sched-ext support
 
 CachyOS provides kernels, which have OOB support for the sched-ext framework.
 Following kernels are supported:
+
 - linux-cachyos (default kernel)
 - linux-cachyos-sched-ext (latest Stable release)
 - linux-cachyos-sched-ext-debug (This is mainly for developers to develop and work on sched-ext)
```
1Naim:
```suggestion
- linux-cachyos-sched-ext-dbg (This is mainly for developers to develop and work on sched-ext)
```

src/content/docs/configuration/sched-ext.md
```diff
@@ -9,16 +9,18 @@ build another kernel for a different scheduler.
...
 - linux-cachyos-rc (latest testing release with the latest features)
 
-You can simply check with following command, if your kernel supports sched-ext:
+You can simply check if your kernel supports sched-ext with the following command:
+
 ```bash
 ❯ zcat /proc/config.gz | grep SCHED_CLASS_EXT
 CONFIG_SCHED_CLASS_EXT=y
```
1Naim:
```suggestion
cat /sys/kernel/sched_ext/state
```

Do you think its better to check for this sysfs file instead of quering the kernel config? Having `CONFIG_SCHED_CLASS_EXT` enabled doesn't necessarily mean sched-ext support is enabled because it also depends on other kernel configs, e.g. `DEBUG_BTF` and some pahole configs. cc @ptr1337

src/content/docs/configuration/sched-ext.md
```diff
@@ -70,50 +74,147 @@ sudo systemctl start scx
 sudo systemctl stop scx
 ```
 
-## Brief introduction to the main ones
+### Adding flags to the scheduler
 
-Since there are many schedulers to choose from, we want to give a little introduction about the schedulers in hand:
+You can include any flag to the sched-ext schedulers. Check out this [guide.](<https://github.com/sched-ext/scx/blob/main/services/README.md>)
```
1Naim: We should put this as a link at the bottom of the page under "See More". This way, we can also add other references, e.g. andrea's and changwoo's blog posts.

src/content/docs/configuration/sched-ext.md
```diff
@@ -70,50 +74,147 @@ sudo systemctl start scx
...
+- Multimedia/Audio production (Thanks to the low latency mode)
+- Great interactivity under intensive workloads
+- Power saving
 
-It depends but for mixed workloads meaning it could vary from gaming, programming, video editing, browsing etc. Rusty/Bpfland/Rustland or even LAVD.
+## Scheduler Profiles
 
-Gaming? then you'll have to choose what matters the most for you.
+:::note
+These profiles are only accessible by using [CachyOS Kernel Manager](/configuration/kernel-manager#sched-ext-gui):
```
1Naim: We should move this to kernel manager page instead, its more suitable as these presets are introduced from that program. 

The note is also quite vague, as while it is true these presets are a feature that was introduced from CachyOS Kernel Manager, it is also possible to replicate these presets by setting scx scheduler flags manually as this is what kernel manager does.

src/content/docs/configuration/sched-ext.md
```diff
@@ -70,50 +74,147 @@ sudo systemctl start scx
...
+
+## General recommendations
+
+### LAVD Autopilot & Autopower
+
+***Quotes from Changwoo Min:***
+In the autopilot mode, the scheduler dynamically changes its power mode (Powersave, Balanced or Performance) according to system's load (CPU
+utilization).
+
+Autopower: Automatically decide the scheduler's power mode based on the system's energy profile aka ***(EPP: Energy Performance Profile)***.
```
1Naim:
```suggestion
Autopower: Automatically decide the scheduler's power mode based on the system's energy profile aka EPP (Energy Performance Preference).
```

1Naim (commented):
I have some suggestions that can be made but I won't review to request changes as there are also discussion based comments here.

src/content/docs/configuration/sched-ext.md
Aarrayy: Is to follow markdown guidelines

src/content/docs/configuration/sched-ext.md
Aarrayy: Hm, but using that command is for checking if sched-ext is currently being used or not instead of: Does my kernel have support for it?.

1Naim: It also works for the latter, since if the kernel doesn't have sched-ext support, the sysfs directory would **not** be there.

1Naim (approved):
Gut

PR #63, opened 2024-09-06
configuration/secure_boot_setup: Remove migration guide and update image signing note

1Naim:
Depends on https://github.com/CachyOS/CachyOS-Settings/pull/81


1Naim:
Drafting because I think the note may be a bit confusing so I want to rework it.

Aarrayy (approved):
Wording & Explanations are fine and easy to understand.

PR #65, opened 2024-09-09
chore: Move shell configuration to post-install

1Naim:


Aarrayy:
I'm 50/50 on this one, maybe its better to rename it instead of moving?

1Naim:
> I'm 50/50 on this one, maybe its better to rename it instead of moving?

to what?

Currently the page doesn't have enough content to warrant its own page, nor is it of high importance. Usually those who really want to change their shell already knows what to do. It is just mainly for an FYI. I think it makes sense to move it here, but I'm open for further discussion.

Aarrayy:
Hmm post setup its a tad empty right now, yeah let's move this to there xD

Aarrayy (approved):

ptr1337 (approved):

PR #69, opened 2024-09-28
features/chwd: Use global autoconfig and sync commands to latest version

1Naim:


src/content/docs/features/chwd.md
```diff
@@ -17,15 +17,17 @@ to use it post-install.
 We can get class IDs by running `lspci`
```
vnepogodin: need to replace `lspci` here then

1Naim: Oops.

vnepogodin (approved):
LGTM

src/content/docs/features/chwd.md
```diff
@@ -14,18 +14,20 @@ to use it post-install.
 
 ### Auto Configuration via Class IDs
 
-We can get class IDs by running `lspci`
+We can get class IDs by running the following command
```
ventureoo: What is the point to run chwd with explicit set class id instead of autoconfiguring all devices with any class id? I mean, the average user just wants all their drivers to be installed properly.

1Naim: I wasn't aware that -a runs without accepting any additional arguments. Just noticed that theres a bracket wrapped around `<classid>`. I assume that implies explicit class id is optional. I'll update accordingly then

ventureoo (approved):
LGTM

PR #71, opened 2024-10-13
installation/installation_on_root: Add dual boot setup

1Naim:
On systemd-boot and GRUB, there needs to be additional setup before both boot managers to detect Windows and add it as a
valid boot entry. For systemd-boot, manually copying over Windows's EFI folders to `/boot/EFI` is necessary while we need
to install and enable `os-prober` with GRUB. As far as I'm aware, rEFInd has full auto-detection for this so no additional
setup is necessary.

I've also removed `installation_dualboot` because that page only has a link to a youtube video on it, and moved that video
link to `installation_on_root` instead.

Besides adding a guide for dual boot, this PR also cleans up the wiki a bit by removing the virtual machine setups for VirtualBox and VMware (refer to commit message for rationale).


ptr1337 (approved):

Aarrayy (approved):
Everything looks great.

PR #73, opened 2024-11-12
Fix typo in link pointing to a github issue

kostadinsh:
Fix typo in link, so it actually opens the issue instead of starting a search.


1Naim (approved):
Thanks!

PR #75, opened 2024-11-21
Remove Dangerous Recommendation from the FAQ

RayJW:
This gets rid of the `sudo pacman -Sy` instruction in the FAQ. This is a dangerous operation and should basically never be used on Arch. Recommending this **will** lead to broken installs sooner rather than later. For more information, see [the Arch Wiki](https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported).

> That is why partial upgrades are not supported. Do not use:
> 
> - `pacman -Sy package`
> - `pacman -Sy` followed by `pacman -S package` (Note the absence of `-Su` in the installation of the package.)

### DIFF
--- a/src/content/docs/support/faq.mdx
+++ b/src/content/docs/support/faq.mdx
@@ -92,7 +92,7 @@ sudo pacman -Syuu
 
 ```code
 # Example
-sudo pacman -S telegram-desktop    
+sudo pacman -S telegram-desktop
 resolving dependencies...
 looking for conflicting packages...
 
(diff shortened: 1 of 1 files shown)


src/content/docs/support/faq.mdx
```diff
@@ -110,7 +110,7 @@ error: failed retrieving file 'telegram-desktop-2.5.1-1-x86_64.pkg.tar.zst' from
 This means that your local database is outdated therefore the package you're asking for can't be downloaded. Run the following command to update your local database:
```
1Naim: Can you update this to match the below command then?

Aarrayy:
This is a certified RTFM ocassion, thanks for the insight Ray

src/content/docs/support/faq.mdx
RayJW: I reworded it to reflect what the command does :)

1Naim (approved):

PR #80, opened 2024-12-08
Update boot_managers.md

SoongVilda:
This commit adds a comprehensive feature comparison table for boot managers (systemd-boot, rEFInd, GRUB) to help users effectively choose the right boot manager for their CachyOS installation. The table includes key features, pros and cons, and recommendations for each boot manager. Additional details are provided below the table for further clarification. This update aims to improve user understanding and decision-making while setting up their system.

### DIFF
--- a/src/content/docs/installation/boot_managers.md
+++ b/src/content/docs/installation/boot_managers.md
@@ -18,6 +18,25 @@ handles the task of loading an OS kernel into memory, often along with support f
 For a more detailed explanation and the reason both terms are often used interchangeably,
 please see "[Managing EFI Boot Loaders for Linux: Basic Principles](https://www.rodsbooks.com/efi-bootloaders/principles.html)" by Rod Smith.
 
+Here is the feature comparison table for the boot managers, accompanied by additional details and recommendations:
+
+### **Feature Comparison Table**
+
+| **Feature**                | **systemd-boot**                     | **rEFInd**                            | **GRUB**                           |
+|----------------------------|---------------------------------------|---------------------------------------|------------------------------------|
+| **UEFI Support**            | ✅ Yes                               | ✅ Yes                               | ✅ Yes                            |
+| **BIOS Support**            | ❌ No                                | ❌ No                                | ✅ Yes                            |
+| **Autodetection**           | ❌ Limited (manual configuration)    | ✅ Full autodetection                | ✅ Autodetection (regenerates config) |
+| **Theming/Customization**   | ❌ Minimal                           | ✅ Extensive                          | ✅ Extensive                      |
+| **Configuration Complexity**| Simple                              | Moderate                             | Complex                          |
+| **Secure Boot Support**     | ✅ Yes                               | ✅ Yes                               | ✅ Yes                            |
+| **Filesystem Support**      | FAT, FAT16, FAT32                   | FAT, EXT4, BTRFS                     | Most Linux FS                     |
+| **Encrypted Boot**          | ❌ No                                | ❌ No                                | ✅ Yes (boot partition)           |
+| **Boot Time**               | Fastest                             | Moderate                             | Slowest                          |
+| **Hardware Compatibility**  | UEFI-only                          | UEFI-only                           | UEFI and BIOS                    |
+| **Ease of Setup**           | Moderate                            | Easy                                 | Moderate                        |
+| **Best Use Case**           | UEFI systems, single OS             | UEFI, multi-boot                     | BIOS systems or encrypted boot partitions |
+
 ## systemd-boot
 Part of systemd family, systemd-boot was created to be as simple as possible, therefore it only has support for UEFI based systems. This simple yet efficient design ensures it is reliable and fast. However this comes at the cost of advanced features supported by other boot managers.
 
--- a/src/content/docs/installation/boot_managers.md
+++ b/src/content/docs/installation/boot_managers.md
@@ -32,7 +32,6 @@ Here is the feature comparison table for the boot managers, accompanied by addit
 | **Secure Boot Support**     | ✅ Yes                               | ✅ Yes                               | ✅ Yes                            |
 | **Filesystem Support**      | FAT, FAT16, FAT32                   | FAT, EXT4, BTRFS                     | Most Linux FS                     |
 | **Encrypted Boot**          | ❌ No                                | ❌ No                                | ✅ Yes (boot partition)           |
-| **Boot Time**               | Fastest                             | Moderate                             | Slowest                          |
 | **Hardware Compatibility**  | UEFI-only                          | UEFI-only                           | UEFI and BIOS                    |
 | **Ease of Setup**           | Moderate                            | Easy                                 | Moderate                        |
 | **Best Use Case**           | UEFI systems, single OS             | UEFI, multi-boot                     | BIOS systems or encrypted boot partitions |


src/content/docs/installation/boot_managers.md
```diff
@@ -18,6 +18,25 @@ handles the task of loading an OS kernel into memory, often along with support f
...
 
+Here is the feature comparison table for the boot managers, accompanied by additional details and recommendations:
+
+### **Feature Comparison Table**
+
+| **Feature**                | **systemd-boot**                     | **rEFInd**                            | **GRUB**                           |
+|----------------------------|---------------------------------------|---------------------------------------|------------------------------------|
+| **UEFI Support**            | ✅ Yes                               | ✅ Yes                               | ✅ Yes                            |
+| **BIOS Support**            | ❌ No                                | ❌ No                                | ✅ Yes                            |
+| **Autodetection**           | ❌ Limited (manual configuration)    | ✅ Full autodetection                | ✅ Autodetection (regenerates config) |
```
1Naim, line 29: As a reader I don't know what "autodetection" entails. Is it autodetection of various kernel versions or autodetection of the various operating systems installed on the system?

src/content/docs/installation/boot_managers.md
```diff
@@ -18,6 +18,25 @@ handles the task of loading an OS kernel into memory, often along with support f
...
+| **Theming/Customization**   | ❌ Minimal                           | ✅ Extensive                          | ✅ Extensive                      |
+| **Configuration Complexity**| Simple                              | Moderate                             | Complex                          |
+| **Secure Boot Support**     | ✅ Yes                               | ✅ Yes                               | ✅ Yes                            |
+| **Filesystem Support**      | FAT, FAT16, FAT32                   | FAT, EXT4, BTRFS                     | Most Linux FS                     |
+| **Encrypted Boot**          | ❌ No                                | ❌ No                                | ✅ Yes (boot partition)           |
+| **Boot Time**               | Fastest                             | Moderate                             | Slowest                          |
```
1Naim: The difference isn't really that big of a deal to categorize it as fastest, faster, fast, etc. I would remove this.

src/content/docs/installation/boot_managers.md
```diff
@@ -18,6 +18,25 @@ handles the task of loading an OS kernel into memory, often along with support f
...
+| **Hardware Compatibility**  | UEFI-only                          | UEFI-only                           | UEFI and BIOS                    |
+| **Ease of Setup**           | Moderate                            | Easy                                 | Moderate                        |
```
1Naim, line 36: This can "bait" users into thinking that the other boot managers except rEEFInd is "hard" to setup while it's not true. Can you give cases where GRUB and sd-boot are harder to setup than rEFInd? In the context of CachyOS, systemd-boot is complemented with `systemd-boot-manager` which aids in creating new entries for the each kernel installed/updated and a generic way to configure entries from one file (similar to GRUB).

src/content/docs/installation/boot_managers.md
```diff
@@ -18,6 +18,25 @@ handles the task of loading an OS kernel into memory, often along with support f
...
+|----------------------------|---------------------------------------|---------------------------------------|------------------------------------|
+| **UEFI Support**            | ✅ Yes                               | ✅ Yes                               | ✅ Yes                            |
+| **BIOS Support**            | ❌ No                                | ❌ No                                | ✅ Yes                            |
+| **Autodetection**           | ❌ Limited (manual configuration)    | ✅ Full autodetection                | ✅ Autodetection (regenerates config) |
+| **Theming/Customization**   | ❌ Minimal                           | ✅ Extensive                          | ✅ Extensive                      |
+| **Configuration Complexity**| Simple                              | Moderate                             | Complex                          |
+| **Secure Boot Support**     | ✅ Yes                               | ✅ Yes                               | ✅ Yes                            |
+| **Filesystem Support**      | FAT, FAT16, FAT32                   | FAT, EXT4, BTRFS                     | Most Linux FS                     |
```
1Naim, line 33: While it is true that GRUB has support for loads of filesystems, I believe that the support is strictly read-only. For example, see https://github.com/CachyOS/distribution/issues/59 where GRUB fails to write `grub.env` to a BTRFS filesystem.

src/content/docs/installation/boot_managers.md
```diff
@@ -18,6 +18,25 @@ handles the task of loading an OS kernel into memory, often along with support f
...
+
+### **Feature Comparison Table**
+
+| **Feature**                | **systemd-boot**                     | **rEFInd**                            | **GRUB**                           |
+|----------------------------|---------------------------------------|---------------------------------------|------------------------------------|
+| **UEFI Support**            | ✅ Yes                               | ✅ Yes                               | ✅ Yes                            |
+| **BIOS Support**            | ❌ No                                | ❌ No                                | ✅ Yes                            |
+| **Autodetection**           | ❌ Limited (manual configuration)    | ✅ Full autodetection                | ✅ Autodetection (regenerates config) |
+| **Theming/Customization**   | ❌ Minimal                           | ✅ Extensive                          | ✅ Extensive                      |
+| **Configuration Complexity**| Simple                              | Moderate                             | Complex                          |
```
1Naim, line 31: I don't really agree with this either, but I would need to fact check on rEFind's configuration

1Naim (changes requested):
The table seems to raise more questions than it does answering them, particularly the ambigious and vague terms such as autodetection, ease of setup, configuration complexity

src/content/docs/installation/boot_managers.md
SoongVilda: In this case, it should be an autodetection of kernel versions. I can change that.

src/content/docs/installation/boot_managers.md
SoongVilda: Okay, I will remove that section.

src/content/docs/installation/boot_managers.md
SoongVilda: Personally, I have sometimes struggled to configure and maintain GRUB and ended up with a broken system. That's why I wrote this. SystemD booting is easy via systemd-boot-manager, true.

I would say we can also remove this section, because the diffculty of configure this up is marginal?

src/content/docs/installation/boot_managers.md
SoongVilda: I’m not sure how to solve this at the moment, but I can try investigating which filesystems are specifically supported by GRUB.

src/content/docs/installation/boot_managers.md
SoongVilda: I can look into this, but since I don’t personally use rEFInd, it may take some time.

src/content/docs/installation/boot_managers.md
1Naim: In that case, all of them support autodetection, without any caveats. `systemd-boot-manager` will update entries and create new ones for each kernel version. GRUB also has kernel hooks for this, no need to regenerate the config.

src/content/docs/installation/boot_managers.md
1Naim: It is likely sufficient to add a footnote clarifying that certain filesystems aren't supported (right not those that come to mind are BTRFS and ZFS).

src/content/docs/installation/boot_managers.md
1Naim: > I would say we can also remove this section

Instead of removing it altogether, I think replacing it with "Partition Table" would be better. In the context of setting up boot managers, all of it is handled by the installer. The only case where a user would have to do it themselves is through manual partitioning, in which case adding a partition table comparison would make sense. Consider https://wiki.cachyos.org/installation/installation_on_root/#manual-partitioning No. 5 for that. (Note that these partition tables aren't strict requirements[1][2], it's just what CachyOS uses.)

[1] rEFInd used to have a three-partition table, `/boot`, `/boot/efi` and `/`
[2] It is also possible for `systemd-boot` to have separate ESP (EFI binaries) and boot partitions (kernel image and initramfs).

src/content/docs/installation/boot_managers.md
SoongVilda: Done

src/content/docs/installation/boot_managers.md
SoongVilda: Is that better?
```
| **Autodetection of kernel versions** | ✅ Supported                        | ✅ Supported                         | ✅ Supported                         |
```

src/content/docs/installation/boot_managers.md
SoongVilda: https://www.gnu.org/software/grub/manual/grub/html_node/Features.html?

Support multiple filesystem types transparently, plus a useful explicit blocklist notation. The currently supported filesystem types are Amiga Fast FileSystem (AFFS), AtheOS fs, BeFS, BtrFS (including raid0, raid1, raid10, gzip and lzo), cpio (little- and big-endian bin, odc and newc variants), Linux ext2/ext3/ext4, DOS FAT12/FAT16/FAT32, exFAT, F2FS, HFS, HFS+, ISO9660 (including Joliet, Rock-ridge and multi-chunk files), JFS, Minix fs (versions 1, 2 and 3), nilfs2, NTFS (including compression), ReiserFS, ROMFS, Amiga Smart FileSystem (SFS), Squash4, tar, UDF, BSD UFS/UFS2, XFS, and ZFS (including lzjb, gzip, zle, mirror, stripe, raidz1/2/3 and encryption in AES-CCM and AES-GCM).

btw: BTRFS should works, idk what happened.

src/content/docs/installation/boot_managers.md
1Naim: I would just remove this. All of them support it and it seems to be a nobrainer feature to have in any boot manager that users shouldn't even think about this.

src/content/docs/installation/boot_managers.md
1Naim: As I mentioned previously, the support mentioned here seems to indicate a sufficient condition for reading, not writing. The issue seems to be that GRUB doesn't support sparse files, which BTRFS and some others do from compression or so. 

edit: The "unsupported" part of this has to do with saving default entries. If that feature is off, then GRUB is happy.

PR #81, opened 2024-12-10
Add a “note” line at the end

ghost:
Note that CachyOS is not supported on some older configurations or is slow.


Aarrayy:
What makes you think that it would be slow? or not supported? perhaps not expecting a noticeable performance gain on older hardware? i think thats a better way to put it

1Naim:
This is misleading at best. CachyOS will perform just like any other distro if run on old (x86_64v1) hardware.

PR #86, opened 2024-12-31
End-of-Year Wiki Update

1Naim:
I wanted to cover all if not more pages in this PR but it turns out that much more changes are necessary than quick updates here and there so I'll submit this for now. Notably 42f1b4a065edc49d275f204e98e962d3e170c1a4 is incomplete and something that I would like to get back to sometime in the future. 569b7fb0ffd276127e82a16bd57bfb1e630a7ca9 is also something that I want to get in production very soon, so I'll have to get my lazy ass to do this quickly after the new year.

The kernel page received the most changes IMO, and it looks much better now compared to its current iteration so I'm very happy about that.

Last commits on 2024 for the wiki for me :) Happy New Year everyone!


Aarrayy (approved):

1Naim:
Merging this now. A few last impressions I get reading this again is some of the wording can still be improved upon, but that can be left for another day 😃

PR #88, opened 2024-12-31
T2 Installation: Add link to t2linux preinstall guide

AdityaGarg8:
The wiki doesn't seem to have instructions on partitioning and disabling secure boot, which are needed for T2 Macs.

### DIFF
--- a/src/content/docs/installation/installation_t2macbook.mdx
+++ b/src/content/docs/installation/installation_t2macbook.mdx
@@ -20,6 +20,8 @@ See [installation of the firmware](#installation-of-the-firmware)
 Follow the instructions in [Installation Prepare](/installation/installation_prepare#creating-a-bootable-cachyos-usb-drive) on how to download
 the ISO and create a bootable USB drive.
 
+You need to follow additional steps to prepare a T2 MacBook for installing Linux such as partitioning your disk using macOS and disabling secure boot. You can follow the [t2linux Preinstall guide](https://wiki.t2linux.org/guides/preinstall/) for instructions.
+
 Refer to [Installation on Root](/installation/installation_on_root) after creating a bootable USB drive.
 CachyOS applies necessary boot parameters and configurations to the user's T2 MacBook with [CachyOS Hardware Detection](/features/chwd).
 


AdityaGarg8:
Alternatively you probably could take the relevant bits from t2linux guide and add them to your wiki. I'll leave that to you.

1Naim:
> Alternatively you probably could take the relevant bits from t2linux guide and add them to your wiki. I'll leave that to you.

I was gonna bring this up and wonder if it would make sense to port it over to our wiki instead, but then I thought about the percentage of users that have already installed CachyOS on their T2 devices. Are the partitioning steps actually necessary or is it just to make it possible for dual booting?

AdityaGarg8:
macOS uses APFS file system, which only their disk utility can partition. We generally suggest dual booting since macOS acts like a backup if Linux messes up. It's also not easy to bring back macOS once removed. You can read more about this here: https://wiki.t2linux.org/roadmap/#can-i-completely-remove-macos

And as far as secure boot in concerned, the T2 Security chip doesn't allow even shim and other ways for secure boot on Linux, so it has to be disabled.

AdityaGarg8:
If the user still wants to remove macOS, partitioning isn't needed

1Naim:
> It's also not easy to bring back macOS once removed

That's a very good point. With that in mind I think this becomes much more higher priority and should probably be added to the wiki itself rather than a reference.

> And as far as secure boot in concerned, the T2 Security chip doesn't allow even shim and other ways for secure boot on Linux, so it has to be disabled

You need to disable it anyway to boot into the ISO so I'm fine in this regard. What I'm contemplating is whether or not the steps to disable secure boot on a Mac should be added to the wiki too. Are the ways to enter MacOS recovery well known among MacOS users (tinkerers)? If so then I don't think we need to add it and a simple reference would be enough, if not then we should add it too.

AdityaGarg8:
> > It's also not easy to bring back macOS once removed
> 
> That's a very good point. With that in mind I think this becomes much more higher priority and should probably be added to the wiki itself rather than a reference.
> 
> > And as far as secure boot in concerned, the T2 Security chip doesn't allow even shim and other ways for secure boot on Linux, so it has to be disabled
> 
> You need to disable it anyway to boot into the ISO so I'm fine in this regard. What I'm contemplating is whether or not the steps to disable secure boot on a Mac should be added to the wiki too. Are the ways to enter MacOS recovery well known among MacOS users (tinkerers)? If so then I don't think we need to add it and a simple reference would be enough, if not then we should add it too.

Not everyone is aware of steps to disable secure boot. You can also add Apple's official article if you wanna add a reference.

https://support.apple.com/en-in/102522

In case you use this reference, make sure you ask to users to set the secure boot settings to no security and enable booting from external media.

1Naim:
> > > It's also not easy to bring back macOS once removed
> > 
> > 
> > That's a very good point. With that in mind I think this becomes much more higher priority and should probably be added to the wiki itself rather than a reference.
> > > And as far as secure boot in concerned, the T2 Security chip doesn't allow even shim and other ways for secure boot on Linux, so it has to be disabled
> > 
> > 
> > You need to disable it anyway to boot into the ISO so I'm fine in this regard. What I'm contemplating is whether or not the steps to disable secure boot on a Mac should be added to the wiki too. Are the ways to enter MacOS recovery well known among MacOS users (tinkerers)? If so then I don't think we need to add it and a simple reference would be enough, if not then we should add it too.
> 
> Not everyone is aware of steps to disable secure boot. You can also add Apple's official article if you wanna add a reference.
> 
> https://support.apple.com/en-in/102522
> 
> In case you use this reference, make sure you ask to users to set the secure boot settings to no security and enable booting from external media.

Alright then, let's add it to the wiki. Can you take care of it? I would be much more comfortable with you doing it since it's your area of expertise but if not I can accept this for now and I can revise it at a later date.

AdityaGarg8:
> > > > It's also not easy to bring back macOS once removed
> > > 
> > > 
> > > That's a very good point. With that in mind I think this becomes much more higher priority and should probably be added to the wiki itself rather than a reference.
> > > > And as far as secure boot in concerned, the T2 Security chip doesn't allow even shim and other ways for secure boot on Linux, so it has to be disabled
> > > 
> > > 
> > > You need to disable it anyway to boot into the ISO so I'm fine in this regard. What I'm contemplating is whether or not the steps to disable secure boot on a Mac should be added to the wiki too. Are the ways to enter MacOS recovery well known among MacOS users (tinkerers)? If so then I don't think we need to add it and a simple reference would be enough, if not then we should add it too.
> > 
> > 
> > Not everyone is aware of steps to disable secure boot. You can also add Apple's official article if you wanna add a reference.
> > https://support.apple.com/en-in/102522
> > In case you use this reference, make sure you ask to users to set the secure boot settings to no security and enable booting from external media.
> 
> Alright then, let's add it to the wiki. Can you take care of it? I would be much more comfortable with you doing it since it's your area of expertise but if not I can accept this for now and I can revise it at a later date.

I don't really have much time for write all of this (no offense). I'll happily review any change you make though.

1Naim:
> > > > > It's also not easy to bring back macOS once removed
> > > > 
> > > > 
> > > > That's a very good point. With that in mind I think this becomes much more higher priority and should probably be added to the wiki itself rather than a reference.
> > > > > And as far as secure boot in concerned, the T2 Security chip doesn't allow even shim and other ways for secure boot on Linux, so it has to be disabled
> > > > 
> > > > 
> > > > You need to disable it anyway to boot into the ISO so I'm fine in this regard. What I'm contemplating is whether or not the steps to disable secure boot on a Mac should be added to the wiki too. Are the ways to enter MacOS recovery well known among MacOS users (tinkerers)? If so then I don't think we need to add it and a simple reference would be enough, if not then we should add it too.
> > > 
> > > 
> > > Not everyone is aware of steps to disable secure boot. You can also add Apple's official article if you wanna add a reference.
> > > https://support.apple.com/en-in/102522
> > > In case you use this reference, make sure you ask to users to set the secure boot settings to no security and enable booting from external media.
> > 
> > 
> > Alright then, let's add it to the wiki. Can you take care of it? I would be much more comfortable with you doing it since it's your area of expertise but if not I can accept this for now and I can revise it at a later date.
> 
> I don't really have much time for write all of this (no offense). I'll happily review any change you make though.

All fine :) Reviewing is plenty! I'll go ahead and merge this for now then. Let's hope users just don't gloss over this one-liner x)

AdityaGarg8:
> > > > > > It's also not easy to bring back macOS once removed
> > > > > 
> > > > > 
> > > > > That's a very good point. With that in mind I think this becomes much more higher priority and should probably be added to the wiki itself rather than a reference.
> > > > > > And as far as secure boot in concerned, the T2 Security chip doesn't allow even shim and other ways for secure boot on Linux, so it has to be disabled
> > > > > 
> > > > > 
> > > > > You need to disable it anyway to boot into the ISO so I'm fine in this regard. What I'm contemplating is whether or not the steps to disable secure boot on a Mac should be added to the wiki too. Are the ways to enter MacOS recovery well known among MacOS users (tinkerers)? If so then I don't think we need to add it and a simple reference would be enough, if not then we should add it too.
> > > > 
> > > > 
> > > > Not everyone is aware of steps to disable secure boot. You can also add Apple's official article if you wanna add a reference.
> > > > https://support.apple.com/en-in/102522
> > > > In case you use this reference, make sure you ask to users to set the secure boot settings to no security and enable booting from external media.
> > > 
> > > 
> > > Alright then, let's add it to the wiki. Can you take care of it? I would be much more comfortable with you doing it since it's your area of expertise but if not I can accept this for now and I can revise it at a later date.
> > 
> > 
> > I don't really have much time for write all of this (no offense). I'll happily review any change you make though.
> 
> All fine :) Reviewing is plenty! I'll go ahead and merge this for now then. Let's hope users just don't gloss over this one-liner x)

Honestly speaking users do gloss over. Has happened many times in our wiki as well!

PR #97, opened 2025-02-09
Add way to get wifi on the live iso in case of T2 Macs

AdityaGarg8:


AdityaGarg8:
Tested on MacBook Pro 16 inch 2019 with the latest CachyOS ISO.

AdityaGarg8:
You may remove the iwd step in case CachyOS makes iwd as the default. wpa_supplicant no longer works with broadcom chips as described here: https://lists.infradead.org/pipermail/hostap/2024-August/042893.html

1Naim:
> You may remove the iwd step in case CachyOS makes iwd as the default. wpa_supplicant no longer works with broadcom chips as described here: https://lists.infradead.org/pipermail/hostap/2024-August/042893.html

It's not, I think we would need to add a chwd profile for broadcom chips to use iwd instead. 

+ @ventureoo @vnepogodin @ptr1337

AdityaGarg8:
> > You may remove the iwd step in case CachyOS makes iwd as the default. wpa_supplicant no longer works with broadcom chips as described here: https://lists.infradead.org/pipermail/hostap/2024-August/042893.html
> 
> It's not, I think we would need to add a chwd profile for broadcom chips to use iwd instead.
> 
> * @ventureoo @vnepogodin @ptr1337

chwd profile I guess applies only after the installation is done, so step 3 remains then I guess.

AdityaGarg8:
In case you compile your own wpa_supplicant, then just revert this patch: https://w1.fi/cgit/hostap/commit/?id=41638606054a09867fe3f9a2b5523aa4678cbfa5

Fedora has already done this.

ventureoo:
> > You may remove the iwd step in case CachyOS makes iwd as the default. wpa_supplicant no longer works with broadcom chips as described here: https://lists.infradead.org/pipermail/hostap/2024-August/042893.html
> 
> It's not, I think we would need to add a chwd profile for broadcom chips to use iwd instead.
> 
>     * @ventureoo @vnepogodin @ptr1337

brcmfmac driver is part of the kernel. I don't think we really want to make a profile that doesn't install anything. It's easier to just set iwd as default in NM.

AdityaGarg8:
Hi

Another workaround to use wifi with wpa_supplicant is adding `brcmfmac.feature_disable=0x82000` to the kernel command line. This way you can still ensure wpa_supplicant is being used.

src/content/docs/installation/installation_t2macbook.mdx
```diff
@@ -12,8 +12,37 @@ import { Tabs, TabItem, Steps } from '@astrojs/starlight/components';
...
+
+1. First boot into macOS and run this command:
+    ```sh
+    curl -sL https://wiki.t2linux.org/tools/firmware.sh | bash -s copy_to_efi
+    ```
+
+2. Now boot into the CachyOS ISO. Open Konsole there, and run these commands:
+    ```sh
+    sudo mkdir -p /tmp/apple-wifi-efi
+    sudo mount /dev/nvme0n1p1 /tmp/apple-wifi-efi
```
1Naim, line 29: Is it always `/dev/nvme0n1p1`? I think it would be better to add an additional step to check for the correct partition.

AdityaGarg8: It's always nvme0n1p1. In fact, the UUID of the EFI partition on T2 Macs is also the same, I.e. `5766-17ED`

AdityaGarg8: What other case were you expecting btw?

1Naim: Cases where `nvme0n1p1` is not the EFI partition on T2 Macs. Can you add a comment there? Something like below should be sufficient

```sh
# Mount MacOS's EFI partition to the created temporary directory. It is commonly /dev/nvm0n1p1
sudo mount /dev/nvme0n1p1 /tmp/apple-wifi-efi
```

AdityaGarg8: > Cases where `nvme0n1p1` is not the EFI partition on T2 Macs. Can you add a comment there? Something like below should be sufficient
> 
> ```shell
> # Mount MacOS's EFI partition to the created temporary directory. It is commonly /dev/nvm0n1p1
> sudo mount /dev/nvme0n1p1 /tmp/apple-wifi-efi
> ```

That's what I was asking. What cases you have in mind? The EFI partition is always the first partition of the SSD. The macOS part of the script will also mount nvme0n1p1 equivalent on macOS and dump the firmware there. If it's missing, it will simply fail.

In some cases there are people who due to lack of experience destroy their Efi partition, and in such cases this whole part is useless. Ideally the script and the APFS driver should be shipped with cachyos, which is the best and full proof way to get the firmware. Although it's completely legal to do so, @ptr1337 still had doubts about that, so this way is the best workaround.

1Naim: > In some cases there are people who due to lack of experience destroy their Efi partition, and in such cases this whole part is useless

Exactly this :) It can be due to lack of experience, or some other insane thought process. Installing Linux on a Mac isn't exactly the most normal thing to do, right? Not sure if you can reinstall MacOS from a completely blank slate, but if it is, it is logically possible that a user can completely wipe their partition table and repartition it in a way that EFI partition is not the first partition.

What I'm trying to say is that we shouldn't make any assumptions wherever possible. An extra sanity check like this is always a good thing :)

AdityaGarg8: > > In some cases there are people who due to lack of experience destroy their Efi partition, and in such cases this whole part is useless
> 
> Exactly this :) It can be due to lack of experience, or some other insane thought process. Installing Linux on a Mac isn't exactly the most normal thing to do, right? Not sure if you can reinstall MacOS from a completely blank slate, but if it is, it is logically possible that a user can completely wipe their partition table and repartition it in a way that EFI partition is not the first partition.
> 
> What I'm trying to say is that we shouldn't make any assumptions wherever possible. An extra sanity check like this is always a good thing :)

I agree, but the thing is even if you do those checks in the Linux bit, the macOS bit will still fail. Probably just say that if this way fails, use a LAN.

AdityaGarg8:
Btw, I also am inclined to remove the firmware package link and just link the wiki page in this guide. Tbh, the link to the package is more illegal that putting an open source APFS driver to the kernel.

AdityaGarg8:
I myself avoid adding any illegal links on the wiki page of t2linux.

src/content/docs/installation/installation_t2macbook.mdx
1Naim: Alright, I'll concede.

AdityaGarg8:
You also could modify your installer to see if any files by the name of \*apple\* exist in `/lib/firmware/brcm` during installation and make sure they make sure they get copied to the final install. This way, the user doesn't have to install the firmware again after installing if they copied the firmware for live install.

PR #102, opened 2025-04-02
wiki: add limine to boot_managers

azdanov:
Hey. I wanted to add some info to the wiki about the new bootloader. Please correct anything that's incorrect, since my knowledge about the topic is very limited :slightly_smiling_face:

### DIFF
--- a/src/content/docs/installation/boot_managers.md
+++ b/src/content/docs/installation/boot_managers.md
@@ -67,20 +67,24 @@ The following is a list of its main pros and cons.
 GRUB is the only available boot loader that supports BIOS booting. It is also the only boot manager that supports boot partition encryption (Different from disk encryption).
 
 ## Limine
+
 Limine is a modern, advanced, and portable multiprotocol bootloader. It serves as the reference implementation for the Limine boot protocol and supports booting Linux as well as chainloading other boot loaders.
 
 ### Pros
+
 - Supports multiple boot protocols, including Multiboot2 and the Linux boot protocols.
 - It can boot on both UEFI and BIOS systems, making it versatile for different hardware configurations.
-- Has theming similar to GRUB and direct support for Btrfs snapshots, which is enabled by default for installations using Btrfs as a filesystem.
+- Has theming capabilities similar to GRUB.
+- Direct support for Btrfs snapshots, which is enabled by default for installations using Btrfs as a filesystem.
 
 ### Cons
-- Only supports a few filesystems, such as FAT12, FAT16, FAT32, and ISO9660, which may require additional setup for systems using other filesystems.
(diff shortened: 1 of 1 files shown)


src/content/docs/installation/boot_managers.md
```diff
@@ -66,6 +66,21 @@ The following is a list of its main pros and cons.
 ### Recommendation
 GRUB is the only available boot loader that supports BIOS booting. It is also the only boot manager that supports boot partition encryption (Different from disk encryption).
 
+## Limine
+Limine is a modern, advanced, and portable multiprotocol bootloader. It serves as the reference implementation for the Limine boot protocol and supports booting Linux as well as chainloading other boot loaders.
+
+### Pros
+- Supports multiple boot protocols, including Multiboot2 and the Linux boot protocols.
+- It can boot on both UEFI and BIOS systems, making it versatile for different hardware configurations.
+- Has theming similar to GRUB and direct support for Btrfs snapshots, which is enabled by default for installations using Btrfs as a filesystem.
```
ptr1337: Lets make the snapshot support as an own point

src/content/docs/installation/boot_managers.md
```diff
@@ -66,6 +66,21 @@ The following is a list of its main pros and cons.
...
+
+### Cons
+- Only supports a few filesystems, such as FAT12, FAT16, FAT32, and ISO9660, which may require additional setup for systems using other filesystems.
```
ptr1337: Maybe mentioned, this only applies to the /boot partition, since it could confuse users

src/content/docs/installation/boot_managers.md
```diff
@@ -66,6 +66,21 @@ The following is a list of its main pros and cons.
...
+- Unlike some other boot loaders, Limine does not automatically add an entry to the NVRAM on UEFI systems; this must be done manually using tools like `efibootmgr`.
```
ptr1337: This can be handled via limine-entry-tool, which is preinstalled out of the box at us

src/content/docs/installation/boot_managers.md
```diff
@@ -66,6 +66,21 @@ The following is a list of its main pros and cons.
...
+
+### Recommendation
+Limine is recommended for users who need a lightweight and versatile bootloader that supports both UEFI and BIOS systems. It is particularly suitable for those who prefer a simple setup with theming options and Btrfs snapshot support.
```
ptr1337: I think we could mention there, that it is also a replacement for grub.
Actually grub is currently barely updated, and due its efi/filesystem drivers it had a lot of security issues

ptr1337 (changes requested):
Made some minor comments

ptr1337 (approved):

PR #106, opened 2025-04-08
fix: improve expressive code theme rendering

azdanov:
I've tried to fix the expressive code theme rendering, since it was broken for some reason. I think it's working much better now. I also updated some dependencies along the way, not all of them, to reduce the risk of breaking.

Screenshot_20250408_122941
Screenshot_20250408_123309
image


vnepogodin (approved):
Many thanks!

PR #109, opened 2025-04-12
chore: fix typescript

azdanov:
I noticed that TypeScript wasn't being checked, so I enabled it. Now, it should properly warn of improper type usage.

I had to fix some issues along the way to make the TypeScript compiler/linter happy. And ran prettier format also.

And I had to rename the ImageComponent import because, for some reason, the linter saw it as a duplicate import. This is probably something to do with how Astro functions internally. There was no issue with MultipleImageComponent, since the Astro file has a different name, with images instead of image.


vnepogodin (approved):
LGTM

PR #113, opened 2025-04-13
revert: fix typescript

azdanov:
This seems to be the easiest way, just to revert my fix.

I tested locally, but could someone confirm also that it works on their machine :slightly_smiling_face:


vnepogodin (approved):
Fair

azdanov:
> Fair

I could have a look some time later on how to properly fix it. But not soon enough.

PR #117, opened 2025-04-19
Document the dlss-swapper script

v1993:
Haven't seen this nice new addition documented anywhere so decided to add a section on it to the wiki.

### DIFF
--- a/src/content/docs/configuration/gaming.mdx
+++ b/src/content/docs/configuration/gaming.mdx
@@ -375,3 +375,58 @@ The Arch Wiki has already provides comprehensive instructions on how to enable [
 - [Ray tracing on Intel](https://wiki.archlinux.org/title/Hardware_raytracing#Intel)
 
 **With this guide, you are now ready to start gaming on CachyOS. Enjoy!**
+
+## Forcing the latest DLSS preset
+
+CachyOS provides a wrapper script `dlss-swapper` for making games supporting DLSS 2 and later utilize the latest preset, improving visual clarity (but potentially reducing performance, especially on older cards).
+
+If `dlss-swapper` is not working or causing issues try updating game's DLSS implementation manually by replacing `nvngx_dlss.dll` with an up-to-date version and using the `dlss-swapper-dll` wrapper script instead.
+
+### How to add dlss-swapper to Steam, Lutris and Heroic Games Launcher
+
+<Tabs>
+
+<TabItem label='Steam'>
+
+<Steps>
+1. Open your `Steam Library`
+2. Right click the game's title and select `Properties`.
+3. On the `General` tab you'll find `Launch Options` section.
+4. Add the following Launch Option:
+    ```sh
+    dlss-swapper %command%
+    ```
+</Steps>
+
+</TabItem>
+
+<TabItem label='Heroic Games Launcher'>
+
+<Steps>
+1. On the left panel open `Settings`.
+2. Go to `Game defaults` then click on `Advanced`.
+3. In the `wrapper` command section. Add the following line without any argument:
+   ```sh
+   dlss-swapper
+   ```
+4. Click on the `+` sign to save changes.
+</Steps>
+
+</TabItem>
+
+<TabItem label='Lutris'>
+
+<Steps>
+1. On the top right open the `hamburger menu`
+2. Go to `Preferences/Global options`.
+3. Enable `Advanced Mode` on the top right.
+4. Scroll down to `Command prefix` and add the following line:
+   ```sh
+   dlss-swapper
+   ```
+5. Save changes.
+</Steps>
+
+</TabItem>
+
+</Tabs>
--- a/src/content/docs/features/cachyos_settings.mdx
+++ b/src/content/docs/features/cachyos_settings.mdx
@@ -62,6 +62,8 @@ Allows access to the `cpu_dma_latency` device node by the audio group.
 - **`cachyos-bugreport.sh`** - Collects various logs from `inxi`, `dmesg` and `journalctl` to aid in troubleshooting
 - **`game-performance`** - Wrapper script for `powerprofilesctl` to switch to performance profile on-demand.
 See [Power Profile Switching On Demand](/configuration/gaming#power-profile-switching-on-demand)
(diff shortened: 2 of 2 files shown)


src/content/docs/configuration/gaming.mdx
```diff
@@ -375,3 +375,58 @@ The Arch Wiki has already provides comprehensive instructions on how to enable [
 - [Ray tracing on Intel](https://wiki.archlinux.org/title/Hardware_raytracing#Intel)
 
 **With this guide, you are now ready to start gaming on CachyOS. Enjoy!**
+
+## Forcing the latest DLSS preset
+
+CachyOS provides a wrapper script `dlss-swapper` for making games supporting DLSS 2 and later utilize the latest preset, improving visual clarity (but potentially reducing performance, especially on older cards).
+
+If `dlss-swapper` is not working or causing issues try updating game's DLSS implementation manually by replacing `nvngx_dlss.dll` with an up-to-date version and using the `dlss-swapper-dll` wrapper script instead.
```
1Naim, line 383:
```suggestion
If `dlss-swapper` is not working or causing issues, try updating game's DLSS implementation manually by replacing `nvngx_dlss.dll` with an up-to-date version and using the `dlss-swapper-dll` wrapper script instead.
```

You mentioned manually replacing the dll file but nowhere in this guide tells you how to do that? Can you add that to the wiki or atleast point to a guide that shows users how to do that?

1Naim (commented):
Can you also add this script to https://wiki.cachyos.org/features/cachyos_settings/#helper-scripts ?

Thanks.

src/content/docs/configuration/gaming.mdx
v1993: I'm a little unsure about this part as it's a bit of an involved process:

1. You need to find the DLL in game files, path to which is going to vary depending on the game in question. Would something like "run `find . -name nvngx_dlss.dll` in the game installation directory to find the exact file location" do?
2. There's no official standalone distribution of the DLL by Nvidia, although I presume [techpowerup](https://www.techpowerup.com/download/nvidia-dlss-dll/) would be reputable enough?

1Naim: > 2\. There's no official standalone distribution of the DLL by Nvidia, although I presume [techpowerup](https://www.techpowerup.com/download/nvidia-dlss-dll/) would be reputable enough?

Yeah, I guess this would be problematic. Do you know how `PROTON_ENABLE_NGX_UPDATER` fetches the latest dll? Additionally, how rare/common is it for that to fail and a manual dll replacement is needed?

v1993: > Do you know how `PROTON_ENABLE_NGX_UPDATER` fetches the latest dll?

No, and it's apparently implemented inside Nvidia's proprietary libraries and possibly loads just models rather than the whole DLL.

> Additionally, how rare/common is it for that to fail and a manual dll replacement is needed?

Seemingly less and less with newer Nvidia drivers; can't say much from my own experience due to having next to no games to test it.

1Naim: Depending on how fast you want to get this merged, I would say to omit this manual dll replacement part and put it as a separate PR due to the concerns above, but if you want to ensure completeness in one PR, we can wait around until it's sorted. Apart from this part, this LGTM.

edit: One more concern below, nothing major.

src/content/docs/features/cachyos_settings.mdx
```diff
@@ -62,6 +62,8 @@ Allows access to the `cpu_dma_latency` device node by the audio group.
 - **`cachyos-bugreport.sh`** - Collects various logs from `inxi`, `dmesg` and `journalctl` to aid in troubleshooting
 - **`game-performance`** - Wrapper script for `powerprofilesctl` to switch to performance profile on-demand.
 See [Power Profile Switching On Demand](/configuration/gaming#power-profile-switching-on-demand)
+- **`dlss-swapper`** - Wrapper script for forcing the latest DLSS preset in games that support the technology
+- - **`dlss-swapper-dll`** - Like `dlss-swapper`, but requires manually updating the `nvngx_dlss.dll` library shipped with the game; may work with games that have issues with the regular version of the script
```
1Naim, line 66: I would say this is a bit misleading, no? (i'm clueless, knowledge is based on looking at the script itself)

Would something like below be more accurate in terms of functionality?

```md
Like `dlss-swapper`, but uses the game's builtin nvngx_dlss instead of force updating it to the latest available.
```

v1993: It's... complicated. Rare exceptions notwithstanding (e.g. the launch version of The Talos Principle Reawakened, now corrected in a hotfix), a game will already expose whichever latest DLSS presets are present in the version of DLL it's shipping with, making `dlss-swapper-dll` mostly useless unless the DLL is updated manually. So... your description is technically correct (with one nitpick that the normal version's NGX updater doesn't touch the DLL itself), but doesn't really explain how you're expected to use the script or the practical implications of using game's built-in `nvngx_dlss.dll` (which are that the script is likely to do nothing by itself).

src/content/docs/configuration/gaming.mdx
ventureoo: AFAIK, in Linux the DLL is part of the nvidia-utils package at the path ``/usr/lib/nvidia/wine/nvngx_dlssg.dll``.

See https://download.nvidia.com/XFree86/Linux-x86_64/575.51.02/README/ngx.html.

1Naim: Yep. More specifically, it started being shipped in the driver from 570 release.

v1993: `nvngx_dlssg` seems to be a separate library from `nvngx_dlss` (and then there's also `nvngx_dlssd`, but that one might just include debugging symbols?); for starters, their sizes are 7.3 MB vs 47 MB.

src/content/docs/features/cachyos_settings.mdx
1Naim: Okay, let's keep it as-is.

1Naim:
@ventureoo @ptr1337 WDYT, is this good to merge?

ptr1337 (approved):

PR #122, opened 2025-04-26
feat(configuration): add wireless-regdb to post install setup

### LINKED ISSUE #93: Explain wireless-regdom configuration
> This configures the connection to use appropriate channels and unlocks additional channels, potentially improving internet speed
>Note: A generic region is set by default; customizing it to your region is recommended for optimal performance

This is an excerpt from the latest update announcement. I think it would be nice to add it (can be as is) to the wiki, because, personally, I wouldn't be aware of this if I didn't read through the announcement.

azdanov:
Add some info about configuring Wi-Fi `wireless-regdb`.

Closes #93

### DIFF
--- a/src/content/docs/configuration/post_install_setup.mdx
+++ b/src/content/docs/configuration/post_install_setup.mdx
@@ -251,7 +251,7 @@ The `wireless-regdb` package includes a **database of wireless rules** (allowed 
    sudo micro /etc/conf.d/wireless-regdom
    ```
 2. **Set your country:**
-   Uncomment the line with your **two-letter ISO country code** (e.g., `WIRELESS_REGDOM="US"`). Ensure only one country is uncommented.
+   Uncomment the line with your **two-letter [ISO country code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements)** (e.g., `WIRELESS_REGDOM="US"`). Ensure only one country is uncommented.
 3. **Reboot:**
    A reboot is typically required for the change to take effect .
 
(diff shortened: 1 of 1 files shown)


src/content/docs/configuration/post_install_setup.mdx
```diff
@@ -239,6 +239,32 @@ Be careful when configuring firewall rules, as improperly configured rules can l
...
+
+**Verification:**
+
+To **check the currently applied regulatory domain**, use the command :
+
+```bash
+iw reg get
+```
+
+Look for the `country XX:` line, where `XX` should match the code you set. If it shows `country 00:`, the system might be using default restrictions or hasn't yet determined the region.
```
1Naim: Would it be better if you showed the output directly? e.g. if you're setting it to HK

```bash
❯ iw reg get
global
country HK: DFS-ETSI # Country correctly shows as HK
        (2400 - 2483 @ 40), (N/A, 36), (N/A)
        (5150 - 5250 @ 80), (N/A, 23), (N/A), NO-OUTDOOR, AUTO-BW
        (5250 - 5350 @ 80), (N/A, 23), (0 ms), NO-OUTDOOR, DFS, AUTO-BW
        (5470 - 5730 @ 160), (N/A, 27), (0 ms), DFS
        (5730 - 5850 @ 80), (N/A, 36), (N/A)
        (5925 - 6425 @ 160), (N/A, 14), (N/A)
```

And so we might want to include an example output of a stock setup with no config too.

src/content/docs/configuration/post_install_setup.mdx
```diff
@@ -239,6 +239,32 @@ Be careful when configuring firewall rules, as improperly configured rules can l
...
+
+**Configuration:**
+
+1. **Edit the configuration:**
+   Open `/etc/conf.d/wireless-regdom` with root privileges .
+   ```bash
+   sudo micro /etc/conf.d/wireless-regdom
+   ```
+2. **Set your country:**
+   Uncomment the line with your **two-letter ISO country code** (e.g., `WIRELESS_REGDOM="US"`). Ensure only one country is uncommented.
```
1Naim: Should probably link to https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements in case anyone doesn't know their two-letter country code for some reason.

azdanov: Good idea. Developer bias is strong with me :smile:

PR #128, opened 2025-05-01
fix(theme): update light theme chevron color to be less bright

azdanov:
image
vs
image

### DIFF
--- a/src/tailwind.css
+++ b/src/tailwind.css
@@ -136,7 +136,7 @@
 
   /* Arrows */
 [data-theme="light"] sl-sidebar-state-persist > ul > li > details { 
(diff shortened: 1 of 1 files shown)


1Naim:
I'm confused, which one is before and after (this is probably useless, i'm colorblind and can't tell the difference 😂 )

azdanov:
Top one is new, it makes the chevron darker. Bottom one is current, it is too bright (at least to my eyes), since it's using the same color as in the dark theme mode.

1Naim:
Oh, so that's what chevron is :)

1Naim (approved):

PR #133, opened 2025-05-13
feat(support): add section about slow applications due to kwallet to faq

azdanov:
Noticed that this question keeps repeated in chats. So added some info about it.

### DIFF
--- a/src/content/docs/support/faq.mdx
+++ b/src/content/docs/support/faq.mdx
@@ -170,6 +170,22 @@ sudo rm -r /var/cache/pacman/pkg/*
 sudo pacman -S nvidia
 ```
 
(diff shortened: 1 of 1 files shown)


1Naim:
Isn't the alleged slowness here caused by the recent update? If so, then the nature of this is very temporary, and I'd like to NACK this.

azdanov:
> Isn't the alleged slowness here caused by the recent update? If so, then the nature of this is very temporary, and I'd like to NACK this.

Good point. I didn't know it was temporary. Never had this issue myself.

TheWyn:
Can be followed here https://bugs.kde.org/show_bug.cgi?id=504014 @azdanov

PR #143, opened 2025-06-24
Fix pacman backup creation

wereii:


src/content/docs/features/optimized_repos.mdx
```diff
@@ -174,7 +174,7 @@ sudo ./cachyos-repo.sh --remove
    ```
 3. Create a backup of the pacman config file:
    ```sh
-   sudo mv /etc/pacman.conf.bak /etc/pacman.conf
+   sudo cp /etc/pacman.conf /etc/pacman.conf.bak
```
1Naim: This is the wrong fix unfortunately. The command is correct, but the step statement is wrong. Instead of create a backup, it should be restore the backup. Can you fix that instead?

wereii: Changed

PR #144, opened 2025-07-05
refactor: update dependencies

azdanov:
- **fix lint:eslint issues**
- **update minor versions**
- **run format**
- **update to tailwindcss 4**


azdanov:
@vnepogodin @SoulHarsh007  Still need to clean up some old css stuff, fix odd tsconfig error popping up about `eslint__js` and test that nothing broke (compare with current deployed wiki).

But it seems to be working.

azdanov:
I think it's ready. Please check that it's builds and all is displayed correctly.
It works on my machine :tm:

azdanov:
Fyi https://bun.sh/blog/bun-lock-text-lockfile
That's why I had odd issues with bun

vnepogodin (approved):
Tested locally

LGTM

src/components/Image.tsx
```diff
@@ -1,7 +1,7 @@
 import { PhotoProvider, PhotoView } from 'react-image-previewer';
 import { CloseButton } from 'react-image-previewer/ui';
 
-const ImageComponent = ({ imgsrc, alt }: { imgsrc: string; alt?: string }) => {
+const Image = ({ imgsrc, alt }: { imgsrc?: string; alt?: string }) => {
```
azdanov, line 4: I had to rename because astro implicitly creates components, so there were 2 `image-component` one from `.tsx` another from `.astro`

src/env.d.ts
```diff
@@ -1,2 +1 @@
-/// <reference path="../.astro/types.d.ts" />
-/// <reference types="astro/client" />
+import 'astro/astro-jsx';
```
azdanov, line 1: I don't think even this is needed. Maybe need to delete the file, since it's not necessary.

The previous references that I removed are automatically included by tsconfig I think.

tailwind.config.cjs
```diff
@@ -1,27 +0,0 @@
-const starlightPlugin = require('@astrojs/starlight-tailwind');
```
azdanov, line 1: config is no longer needed for tailwindcss v4

tsconfig.json
```diff
@@ -1,13 +1,31 @@
 {
   "extends": "astro/tsconfigs/strictest",
   "compilerOptions": {
-    "jsx": "react-jsx",
+    "plugins": [
+      {
+        "name": "@astrojs/ts-plugin"
```
azdanov, line 6: This is used for `astro check` command

src/components/Image.tsx
vnepogodin: yeah due that reason it was named Component. it's fine

SoulHarsh007 (approved):
LGTM!

PR #147, opened 2025-07-21
fix: installation instructions for bottles 

netrolite:
The page at https://wiki.cachyos.org/configuration/gaming says that to install bottles you need to run `sudo pacman -S bottles`, which doesn't actually work because bottles isn't in the arch repositories.

The Bottles documentation [says](https://docs.usebottles.com/getting-started/installation) the only supported way of installing bottles is through flatpak, so this PR adds a command for installing flatpak (`sudo pacman -S flatpak`) and a command for installing bottles (`flatpak install bottles`)

### DIFF
--- a/src/content/docs/configuration/gaming.mdx
+++ b/src/content/docs/configuration/gaming.mdx
@@ -352,7 +352,7 @@ Now you can install bottles using flatpak:
 sudo flatpak install bottles
 ```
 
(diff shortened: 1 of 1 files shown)


ptr1337 (approved):
Thanks for the PR - seems good.

Even tough, I think on CachyOS we should really prefer using native libaries. Due bottles packaging and software changes this is not working.

Ill merge that for now and will talk in the team, how we should proceed there in the future :)

PR #150, opened 2025-07-23
gaming: remove bottles entry

vnepogodin:
We don't recommend using flatpaks, and rather not mention it on the wiki

As bottles only support flatpak, removing bottles entry entirely


ventureoo (approved):

ptr1337 (approved):

cherrychon:
Well I like Bottles...
But I hate flatpaks w/ a burning passion.
So that's a great thing

PR #151, opened 2025-07-24
feat(desktop_environments): add kbd plugin for keymaps

### LINKED ISSUE #149: Improve the keyboard shortcut display
At the moment, normal text is used to showcase shortcuts. It would be good to use a different method or a plugin, such as https://starlight-kbd.netlify.app/, for visual improvement.

Here's an example of https://wiki.cachyos.org/desktop_environments/hyprland/

Image

azdanov:
Closes: #149 

image

Need to also add this to other translations. Some day.


vnepogodin (approved):
Many thanks. looks sexy

Aarrayy (approved):
Thanks for the suggestion!, looks good

PR #155, opened 2025-07-28
fix(hyprland): update screenshot keybind

azdanov:


azdanov:
@Ly-sec cc

Ly-sec:
> @Ly-sec cc

Looks fine to me. No idea where the Super+A even came from to be honest.

PR #162, opened 2025-08-02
docs(readme): update readme for new translation setup

azdanov:
Forgot to update readme for the new setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,22 +59,19 @@ You can use the localization tracker at [https://wiki.cachyos.org/localization](
 
 ### Getting started
 
-Firstly, you need to add your language to the list of available locales to the site. This is done in `astro.config.mjs`.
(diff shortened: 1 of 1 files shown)


azdanov:
@SoongVilda Could you please check this? Would be nice to have up-to-date readme :slightly_smiling_face:

SoongVilda:
It looks good and intuitive. One thing: Maybe we should add instructions on how to properly test changes? I hit a wall this morning (you know, hah). We use `build`, but our README recommends `bun run dev`.

azdanov:
Yes, `bun run build` would be better. After building, you can run `bun run preview`, and it should show all the changes as in the production environment. One caveat is that the translation tracker expects all changes to be committed when doing a production build, so it will show an error.

azdanov:
@Shendisx cc

Aarrayy (approved):

PR #166, opened 2025-08-04
Add "increasing shader cache size" in gaming section

popcar2:
Adds a note to improve gaming performance by increasing the shader cache size limit. I spent a while figuring out why my games were re-compiling shaders each launch, and increasing the shader cache size fixed it perfectly.

AFAIK this is something that some gaming distros do by default, and is a relatively safe fix.

### DIFF
--- a/src/content/docs/configuration/gaming.mdx
+++ b/src/content/docs/configuration/gaming.mdx
@@ -196,7 +196,8 @@ To avoid long loading times and stuttering, we can increase the global shader ca
 	# Increase Nvidia's shader cache size to 12GB
 	export __GL_SHADER_DISK_CACHE_SIZE=12000000000
 	```
-4. Save the file and restart your device.
+4. Save the file by pressing `CTRL+X`, `Y`, then `Enter`.
+5. Restart your device.
 </Steps>
 
 </TabItem>
@@ -217,7 +218,8 @@ To avoid long loading times and stuttering, we can increase the global shader ca
 	# Increase AMD's shader cache size to 12GB
 	export MESA_SHADER_CACHE_MAX_SIZE=12G
 	```
-4. Save the file and restart your device.
+4. Save the file by pressing `CTRL+X`, `Y`, then `Enter`.
+5. Restart your device.
 </Steps>
 
 </TabItem>
--- a/src/content/docs/configuration/gaming.mdx
+++ b/src/content/docs/configuration/gaming.mdx
@@ -197,7 +197,7 @@ To avoid long loading times and stuttering, we can increase the global shader ca
 	export __GL_SHADER_DISK_CACHE_SIZE=12000000000
 	```
 4. Save the file by pressing `CTRL+X`, `Y`, then `Enter`.
-5. Restart your device.
+5. Restart your system.
 </Steps>
 
 </TabItem>
@@ -219,7 +219,7 @@ To avoid long loading times and stuttering, we can increase the global shader ca
 	export MESA_SHADER_CACHE_MAX_SIZE=12G
 	```
 4. Save the file by pressing `CTRL+X`, `Y`, then `Enter`.
-5. Restart your device.
+5. Restart your system.
 </Steps>
 
 </TabItem>
--- a/src/content/docs/configuration/gaming.mdx
+++ b/src/content/docs/configuration/gaming.mdx
@@ -173,6 +173,59 @@ has similar a behavior but it should **not** be used as CachyOS already ships wi
 
 </Tabs>
 
+### Increase maximum shader cache size
+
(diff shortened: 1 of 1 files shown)


src/content/docs/configuration/gaming.mdx
```diff
@@ -173,6 +173,59 @@ has similar a behavior but it should **not** be used as CachyOS already ships wi
...
+
+<Tabs>
+
+<TabItem label='Nvidia GPU'>
+
+<Steps>
+1. Open the terminal.
+2. Enter this command:
+	```sh
+	sudo nano /etc/profile.d/increase-shadercache.sh
```
Aarrayy, line 192: Not all users know how to use nano, perhaps add how to save the file?

Aarrayy (approved):

src/content/docs/configuration/gaming.mdx
```diff
@@ -173,6 +173,61 @@ has similar a behavior but it should **not** be used as CachyOS already ships wi
...
+	```
+3. Paste the following into the file:
+	```sh
+	# Increase Nvidia's shader cache size to 12GB
+	export __GL_SHADER_DISK_CACHE_SIZE=12000000000
+	```
+4. Save the file by pressing `CTRL+X`, `Y`, then `Enter`.
+5. Restart your device.
```
Aarrayy: Device? shouldn't this refer to "Restart your system"?

popcar2: Both words can work but I guess system is more obvious, I'll change it now.

src/content/docs/configuration/gaming.mdx
```diff
@@ -173,6 +173,61 @@ has similar a behavior but it should **not** be used as CachyOS already ships wi
...
+
+<Tabs>
+
+<TabItem label='Nvidia GPU'>
+
+<Steps>
+1. Open the terminal.
+2. Enter this command:
+	```sh
+	sudo nano /etc/profile.d/increase-shadercache.sh
```
ventureoo, line 192: This will not work for GNOME & KDE users if you have run game with app launched from applications menu, since they're all running as systemd user units, that do not inherit environment variables of shell.

popcar2: Hmm, I didn't know about that. The vast majority of people run their games through Steam/Lutris/Heroic though. Is there a more universal way to set the cache size?

If not, we can maybe add a note that this might not work on games launched straight through the apps menu.

ventureoo: > Hmm, I didn't know about that. The vast majority of people run their games through Steam/Lutris/Heroic though. Is there a more universal way to set the cache size?
> 
> If not, we can maybe add a note that this might not work on games launched straight through the apps menu.

It doesn’t seem to matter because my information is outdated and variables are still being exported. So sorry for noise.

popcar2: Ah, no worries. Good to hear that's not an issue.

ventureoo: So, yeah, it seems that GNOME still requires creating a file in ``/etc/environment.d`` or ``~/.config/environment.d``, which adds variable while KDE (apparently) no longer does.

popcar2: Gnome moment... Do you know a way that would be compatible with all desktop environments? I assumed profile.d was the universal way to do it.

ventureoo: > Gnome moment... Do you know a way that would be compatible with all desktop environments? I assumed profile.d was the universal way to do it.

In fact, the real problem is caused by using fish as a system shell. This doesn’t seem to happen on SDDM as it does export variables from ``/etc/profile`` anyway (https://github.com/sddm/sddm/blob/08a4a77073ffc5566baa54fa5345f20eea999648/data/scripts/wayland-session#L20), but it’s not handled on the GDM side, so we have to create an ``environment.d`` files. Maybe we should just add a warning for fish users, or process it in some way in our settings.

popcar2: At this point I think it's worth creating a new issue for the wiki to see what the best option might be. We could add a note but then we'd have to find out if this only happens on Gnome or not, and which other DEs might have issues setting env variables.

ventureoo: I think it’s best to just use ``/etc/environment`` so as not to complicate things for user. We don’t use this when we want to package something, since PAM dropped support for drop-in configuration files a long time ago, but it’s okay for wiki.

popcar2: Done, the PR for it just got merged :+1:

PR #168, opened 2025-08-04
Make the flag longer so people can actually see it

Revival8697:
Its short length + the color in dark mode (all other commands are in yellow only) make it hard to see.

Screenshot_20250805_040743 cleaned


Aarrayy:
Good point

Aarrayy:
Tomorrow i'll take a look into styling these code blocks to improve the visibility, Thanks for the idea.

Aarrayy:
Fixed after b3af3c1

 It should be way easier to read now, take a look

PR #171, opened 2025-08-06
feat(middleware): add ignore i18n fallback for changelogs

azdanov:
This should remove the missing translation banner displayed for `changelogs` route, since those will always be in English.


src/tailwind.css
```diff
@@ -38,30 +37,29 @@ img {
 
 /*Dark Theme*/
 
-/* Content text */ 
+/* Content text */
 
 [data-theme='dark'] body {
```
azdanov, line 42: Formatted by running `bun run format`

Aarrayy (approved):

PR #174, opened 2025-08-07
Update optimized_repos.mdx

capocasa:
Added an easy low risk way to upgrade a production machine from arch


ptr1337:
Mhm, I do not really think that makes much sense outside for installing our kernel, which then could be only put the "cachyos-v3" repository below

capocasa:
It was a great way to build trust- but you're right, I did start with the kernel, which worked great! Best realtime kernel for arch. I'll add that to the arch realtime wiki as well.

I can rephrase this to emphasize kernel installation.

PR #176, opened 2025-08-09
i18n: add french translation

azdanov:
Added translation for French.
cc @Shendisx


azdanov:
I'll close this for now, since there has been many changes to English. If someone wishes to translate then can pick up any bits and pieces from here.

cc @Shendisx

PR #179, opened 2025-08-10
docs: clarify about VMs and GUI package managers

vnepogodin:
Let's clarify some info about our expectations from general system installation on VMs

Also warn people about using GUI package managers, and their possible issues which they can encounter while not banning entirely such usage

### DIFF
--- a/src/content/docs/installation/installation_prepare.mdx
+++ b/src/content/docs/installation/installation_prepare.mdx
@@ -11,6 +11,10 @@ import ImageComponent from '~/components/image-component.astro';
 Before preparing the installation, make sure that the computer used meets the minimum system requirements in order to run CachyOS. The installer
 uses an online installation process so a stable and relatively fast internet connection is mandatory.
 
+:::note[Virtual Machines]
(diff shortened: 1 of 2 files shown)


vnepogodin:
nit: maybe for pamac its a bit harsh wording as keyring issue seems less dangerous than package manager backend crash ;)

src/content/docs/installation/installation_prepare.mdx
```diff
@@ -11,6 +11,10 @@ import ImageComponent from '~/components/image-component.astro';
 Before preparing the installation, make sure that the computer used meets the minimum system requirements in order to run CachyOS. The installer
 uses an online installation process so a stable and relatively fast internet connection is mandatory.
 
+:::note[Virtual Machines]
+It is not recommended to install CachyOS on a virtual machine. VMs can have issues with incorrect configuration or can be broken entirely. It is recommended to install CachyOS on bare metal.
```
ventureoo, line 15: Maybe we’ll just limit the support to QEMU/KVM? Because most of the issues happen with VirtualBox and VMware.

vnepogodin: Still with QEMU people can get also the configuration incorrectly, but usually people which use QEMU get the configuration right

Aarrayy (approved):

PR #184, opened 2025-08-16
Update gaming.mdx

briandipalma:


Aarrayy:
I don't think this is something that most users will do even if they're gamers. Closing for now

PR #202, opened 2025-09-05
Include UFS not supported with Limine

d3nnis77:
Limine currently does not work with UFS-storage and just shows a black screen instead of booting.

Also helpful would be a section that booting from UFS requires to add the modules ufs and ufshcd-pci in mkinitcpio conf after installation.
But I´m unsure where this would fit best (maybe Filesystems?).


Aarrayy:
Good catch about UFS, regarding the question about a section of booting from UFS. We don't have requests or if at all users that install CachyOS on a UFS drive so for now its not considered, maybe in the future if the demands increases. Thanks!

d3nnis77:
Well, I did install it on Chromebook (that´s how I came across the UFS problem) and it works great. Optimization is especially great on low-performance devices.

But I agree that it´s probably a niche use case.

PR #204, opened 2025-09-07
FAQ: Add solution for invalid signatures (Octopi)

Root-Core:
I often encounter this issue, and neither updating the mirrors nor manually resetting the keyring solves it. The only reliable solution is Octopi (for some reason).

I could also provide a german version.

### DIFF
--- a/src/content/docs/cachyos_basic/faq.mdx
+++ b/src/content/docs/cachyos_basic/faq.mdx
@@ -56,7 +56,20 @@ While mirrors often fix themselves after a short while, if the issue persists, y
 
 <Tabs>
 
-<TabItem label='First Solution: Rate Your Mirrors'>
+<TabItem label='First Solution: Update via Octopi'>
+
+This is the easiest option, but it uses the GUI.
+
+- Open Octopi.
+- Press [CTRL]+[K]: This searches for updates.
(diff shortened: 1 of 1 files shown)


Aarrayy:
Maybe your mirrors were unsynced at that moment and due to the bad timing this happened. Thanks for the suggestion tho, i'll add a third possible solution without Octopi.

Aarrayy:
Added with [f3f3a23](https://github.com/CachyOS/wiki/commit/f3f3a23085914b9538e72ed797535f8ffb05ae53)

Root-Core:
Nope, neither syncing nor rating the mirrors helps. This is the only reliable solution on my system.
I'm not exactly new to Arch, these issues are common in and exclusive to CachyOS though.

BTW, regarding f3f3a23: There isn't an "above", as these are tabbed items.

1Naim:
> Nope, neither syncing nor rating the mirrors helps

It might be unfortunate timing, since what you are proposing is 1:1 with `sudo pacman -Syu`.

Related to this topic, @Shendisx , this third solution should actually be the first. Most of the time these invalid signature errors comes from a combination of improper cache invalidation from mirrors and bad timing. In most cases trying to sync after a few minutes will result in a success and users don't need to perform the alternative solutions mentioned.

PR #205, opened 2025-09-08
create folder to mount to

jameskimmel:
Without an existing folder, I was unable to mount.  This does create a folder and for the mountpoint and deletes it afterwards.


Aarrayy:
Good catch, and thanks for the recommendation

PR #211, opened 2025-09-10
Remove RTC as the tweak was reverted

PavelDobCZ23:
Just cleaning up the settings page as I've noticed there's an entry that no longer applies. See the merge below.
https://github.com/CachyOS/CachyOS-Settings/pull/151/commits


Aarrayy:
Totally forgot about that change, thanks!

PR #214, opened 2025-09-30
Fix systemd-boot is EFI driver capable

l-2-j:
The docs are just fully wrong about systemd-boot not being able to use non-FAT partitions, [has been documented on the Arch wiki](https://wiki.archlinux.org/title/Systemd-boot#Installation_using_XBOOTLDR) and can personally attest to it working on CachyOS too just fine. TLDR; FAT `esp`, copy `efifs` UEFI drivers to it for an `XBOOTLDR` marked `/boot` partition on the same disk.

### DIFF
--- a/src/content/docs/installation/boot_managers.mdx
+++ b/src/content/docs/installation/boot_managers.mdx
@@ -61,8 +61,8 @@ If you encounter such issues, consider using `systemd-boot`.
     </tr>
     <tr>
       <td><strong><code>/boot</code> filesystem support</strong></td>
-      <td>FAT12/16/32 for /efi, with ext4/Btrfs via EFI drivers for /boot (<code>efifs</code>)</td>
-      <td>FAT natively; ext4/Btrfs via EFI drivers (<code>efifs</code>)</td>
+      <td>According to firmware's support (usually FAT12/16/32); More with EFI drivers</td>
+      <td>Firmware's,ext2,ext3,ext4,btrfs,ISO-9660,HFS+, and NTFS; More with EFI drivers</td>
       <td>Broad filesystem support (ext*, Btrfs, XFS, etc.)</td>
       <td>FAT12/16/32, ISO9660 for <code>/boot</code></td>
     </tr>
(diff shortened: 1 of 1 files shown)


src/content/docs/installation/boot_managers.mdx
```diff
@@ -61,7 +61,7 @@ If you encounter such issues, consider using `systemd-boot`.
     </tr>
     <tr>
       <td><strong><code>/boot</code> filesystem support</strong></td>
-      <td>FAT12/16/32 only</td>
+      <td>FAT12/16/32 for /efi, with ext4/Btrfs via EFI drivers for /boot (<code>efifs</code>)</td>
```
1Naim:
```suggestion
      <td>FAT12/16/32 for ESP, with ext4/Btrfs via EFI drivers for XBOOTLDR (<code>efifs</code>)</td>
```
Surely ext4/btrfs are not the only filesystems available in efifs?

src/content/docs/installation/boot_managers.mdx
```diff
@@ -133,8 +133,8 @@ Part of the systemd family, systemd-boot was created to be as simple as possible
 #### Cons
 - No support for BIOS/MBR.
 - Very barebones: no theming or customization.
-- Only able to read boot images on EFI-supported filesystems (FAT12/16/32).
-- Cannot find boot images on partitions other than its own ESP.
+- Requires manually setting up separate `esp` and `/boot` (XBOOTLDR) partitions with EFI drivers to support filesystems beyond the default supported ones (FAT12/16/32).
```
1Naim:
```suggestion
- Requires manually setting up separate `ESP` and XBOOTLDR partitions with EFI drivers to support filesystems beyond the default supported ones (FAT12/16/32).
```
Would also be good if u can add a hyperlink for XBOOTLDR.

1Naim (changes requested):

src/content/docs/installation/boot_managers.mdx
Revival8697: I think it was to copy the rEFInd entry (which is wrong btw).

Just fix it to support whatever via https://wiki.archlinux.org/title/Unified_Extensible_Firmware_Interface#UEFI_drivers.

rEFInd should also be fixed to supports what is in https://www.rodsbooks.com/refind/drivers.html#selecting by default and also whatever via the UEFI drivers.

1Naim (approved):

PR #216, opened 2025-10-04
Add link to .sig file

qoheleth-tech:
Link to sig file would have saved me some time trying to find where it was located.


Aarrayy (approved):
Good idea, dunno why it never went through my mind xD

PR #217, opened 2025-10-05
Adding additional information to Google Chrome hardware acceleration RE: Intel GPU's

Bwitzig:
- Added information for Intel GPU's
- Clarified which requirements are optional and when to install them
- Added info for Lunar lake intel GPU (tested on 258v laptop)
- Added info for pre Lunar Lake GPU (tested on 155H)


Aarrayy (approved):
Thanks for the addition!

PR #239, opened 2025-10-31
Add special steps for MSI secure boot setup

Krzeszny:
Since some MSI mobos don't have setup mode, I've discovered how to do it in that case.

Refer to CachyOS discord for details, but there isn't much more to it.
https://discord.com/channels/862292009423470592/1431703601583095888

Note: I didn't test if the site renders the new image correctly.


Aarrayy (approved):

ptr1337:
Is this safe?

Krzeszny:
> Is this safe?

Well, I did it and I'm alive :3 
(Cachy and Windows are also alive and secured)

By the way, I don't remember if it was necessary to re-enable Provision Factory Default Keys after setting up secure boot, but the worst it could do is prevent boot when secure boot is enabled, which is just as easy to fix. 

I'd be happy to test it again if someone told me how to revert to "unsecured boot" so I can set up secure boot again. Do I just delete the keys and call it a day or should I run some commands?

Also, I have this feedback topic about some other steps https://discord.com/channels/862292009423470592/1432249787314606101

1Naim:
Is it necessary to disable "Provision Factory Default Keys"? My understanding is that with the former enabled + deleting all secure boot variables, its equivalent to setup mode.

Krzeszny:
> Is it necessary to disable "Provision Factory Default Keys"? My understanding is that with the former enabled + deleting all secure boot variables, its equivalent to setup mode.

You see, when you reboot after deleting secure boot variables, it provisions the default keys.

1Naim:
> > Is it necessary to disable "Provision Factory Default Keys"? My understanding is that with the former enabled + deleting all secure boot variables, its equivalent to setup mode.
> 
> You see, when you reboot after deleting secure boot variables, it provisions the default keys.

Is that not setup mode? I guess my understanding is wrong if not.

1Naim (approved):
Anyway, this doesn't look brickable so I'm fine with it. I heard that if you enroll your own keys without vendor keys could brick some motherboards, but the step I wrote here enrolls your own keys with Microsoft keys, which should be sufficient.

... and after looking into `sbctl enroll-keys` again, I see that theres also the `-f` flag which also enrolls the builtin firmware keys. I think we should also add that flag here. It's out of scope for this PR, but if you want you can add that here and explain what the flag does too. It's okay if you don't, I plan on adding it after this recent discovery.

Krzeszny:
> > > Is it necessary to disable "Provision Factory Default Keys"? My understanding is that with the former enabled + deleting all secure boot variables, its equivalent to setup mode.
> > 
> > 
> > You see, when you reboot after deleting secure boot variables, it provisions the default keys.
> 
> Is that not setup mode? I guess my understanding is wrong if not.

Nope, provisioning default keys returns the default keys after you delete them, so you can be in setup mode while in UEFI... until you reboot. That's why disabling that setting is necessary, by my understanding. I'll look for some documentation from MSI but no promises.

Again, if someone tells me how to revert to unsecured boot, I could do more testing, but I did it like I said and it worked.

1Naim:
> Again, if someone tells me how to revert to unsecured boot, I could do more testing, but I did it like I said and it worked.

Deleting the variables again and enabling "Provision ..." should revert it to stock settings. I don't have any guarantees though, so it's your decision if you decide to test this or not.

PR #245, opened 2025-11-17
Integrate KaTeX to render inline & block math in MDX

bouness:
**Summary**
This PR adds KaTeX support to the Cachyos wiki, allowing both inline and block math formulas to be rendered in Markdown files.

**Changes**

* Installed `starlight-katex` library
* Added inline`$...$` and block `$$...$$` math formula
* Added markdown rendering plugin `starlightKatex()` to support math

**Example usage in a markdown file:**
```sh
# inline display
When $a \ne 0$, there are two solutions to $(ax^2 + bx + c = 0)$ and they are:

# block display
$$
x = {-b \pm \sqrt{b^2-4ac} \over 2a}
$$
```
**Result:**
When $a \ne 0$, there are two solutions to $(ax^2 + bx + c = 0)$ and they are:

$$
x = {-b \pm \sqrt{b^2-4ac} \over 2a}
$$

**Why**
Enables technical content to include proper mathematical notation, improving readability for engineering and scientific examples.

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -1,3 +1,4 @@
+import { starlightKatex } from 'starlight-katex';
 import { defineConfig } from 'astro/config';
 import starlight from '@astrojs/starlight';
 import react from '@astrojs/react';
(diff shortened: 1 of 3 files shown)


Aarrayy:
I highly doubt we'll ever include any math equation in our wiki. Thanks for the suggestion but this is not needed atm.

PR #247, opened 2025-11-19
Optimized Repositories: Add note about auto architecture check

### LINKED ISSUE #233: "does not have a valid architecture - linux-cachyos-eevdf"
`[god@god ~]$ sudo pacman -Syu
:: Synchronizing package databases...
 core is up to date
 extra                                                                                 8.4 MiB  17.8 MiB/s 00:00 [###################################################################] 100%
 multilib is up to date
 cachyos-v3 is up to date
 cachyos-core-v3 is up to date
 cachyos-extra-v3 is up to date
 cachyos                                                                             411.3 KiB   249 KiB/s 00:02 [###################################################################] 100%
:: Starting full system upgrade...
error: failed to prepare transaction (package architecture is not valid)
:: package linux-cachyos-eevdf-6.7.9-1-x86_64_v3 does not have a valid architecture
:: package linux-cachyos-eevdf-headers-6.7.9-1-x86_64_v3 does not have a valid architecture`

gwuen:
Resolves: CachyOS/linux-cachyos#233

### DIFF
--- a/src/content/docs/features/optimized_repos.mdx
+++ b/src/content/docs/features/optimized_repos.mdx
@@ -183,6 +183,17 @@ is most optimized for it. It also backs up your old `pacman.conf` for repository
    'https://mirror.cachyos.org/repo/x86_64/cachyos/cachyos-v4-mirrorlist-22-1-any.pkg.tar.zst' \
    'https://mirror.cachyos.org/repo/x86_64/cachyos/pacman-7.0.0.r7.g1f38429-2-x86_64.pkg.tar.zst'
    ```
+   :::note
+   At the moment, only CachyOS's pacman fork is able to recognize and automatically check for the optimized architectures.
+   If you chose to skip installing the customized pacman in this step, pacman has to be configured to recognize the appropriate architecture(s) manually.
+   For example:
+
+   ```ini
+   # /etc/pacman.conf
+   [options]
+   Architecture = auto x86_64_v3 x86_64_v4
+   ```
+   :::
 3. Add the CachyOS repositories to the pacman config file:
    :::note
    These repositories should be put above Arch Linux Repositories.


ptr1337:
Hey,
Thanks for the MR.
It is not really something we do want to support, since there are quite a bunch of pacman features we do relay on. 

We know that few enhanced users utilizing it on their own with blocking our pacman, but it often results into not wanted issues and therefore should be not noted in the wiki

gwuen:
I understand, but then [the following note](https://github.com/CachyOS/wiki/blob/b5f14192e20134e5d7d213b8eec87acda5680b3f/src/content/docs/features/optimized_repos.mdx?plain=1#L138-L140) should be removed as it suggests just that:

> If you want to avoid this, don't add the "cachyos" repository, which contains the customized pacman.

Also see the discussion in CachyOS/linux-cachyos#233.

gwuen:
To be clear, I'm fine with either solution. But as it currently stands, the wiki article suggests not using the forked pacman is a “safe” thing to do, but fails to mention any “not wanted issues” or the additional setup in `/etc/pacman.conf` that is necessary for this.

CatmanIX:
Considering this warning was added in September 2024 (https://github.com/CachyOS/wiki/commit/fbe0c06bba01ef582be6526a917d70ca2e141e39) and includes "We are working with Arch Linux to provide a proper compatibility again", is it possible the problem causing the confusion has already been resolved in the year+ since and the warning should just be removed entirely?

Otherwise if it's still an issue it would help to have some details about how this fork "could lead when using the Arch Linux pacman into warnings" and why _not_ using the base cachyos repository would resolve this considering the only information i've been able to find (primarily the issue cited in this pr) suggests the _opposite_ is true. I'm also thinking that may have been what the warning was trying to say in the first place but was somehow confused in it's wording since the issue with arch-pacman not recognising the extended architecture repos as valid without tweaking the conf kinda seems like it could be this "feature validation feature" it's talking about.

vnepogodin:
The "installed db" feature is the primary warning wiki describes as an "issue". As "base pacman" doesn't recognise that, citing it's unknown entry

The extended architecture detection will not cause any warnings or issues when using neither automatic nor editing pacman.conf manually, similarly to "installed db" warnings. At the same time it would cause the warnings due to optimised repositories having "newer packages" when reverting/removing repositories; in case of using "base pacman" with our optimised repositories requires manual intervention in order to determine and set the correct Architecture values, while on the other hand our pacman prevents installation of the packages with "unsupported architecture"

CatmanIX:
So if i understand correctly, the only problem is that "base pacman" has some issues properly handling the optimised repositories, but there shouldn't be _any_ problems using the cachyos fork, right? So the current suggestion to _remove_ the 'cachyos' repo to prevent these problems is mistaken?

If this is the case i can try to throw together a PR this evening to help clarify the wording of the warning and prevent further misunderstandings, i just want to make sure i'm understanding things correctly now first.

gwuen:
> So the current suggestion to remove the 'cachyos' repo to prevent these problems is mistaken?

No, I don't think it is mistaken. As @vnepogodin just confirmed, the incompatibility is still present. But I do agree the suggestion to not install CachyOS's pacman is out of place, as it only leads to more confusion, requires more setup and apparently is not supported either.

CatmanIX:
>No, I don't think it is mistaken. As @vnepogodin just confirmed, the incompatibility is still present.

Yes, but the incompatibility seems to be between Arch's version of pacman and the cachyos optimised repos, while there's no compatibility issues involving Cachy's version of pacman. If this is the case then suggesting to remove the cachyos repo to _avoid_ compatibility problems _must_ be a mistake as doing so seems to be what actually _causes_ the issue being warned about. Or am i still completely misunderstanding something?


Assuming i _am_ understanding correctly, i've whipped together a replacement warning that hopefully clarifies the problem that can occur and how to properly avoid and resolve it:

> :::caution
> CachyOS provides a modified fork of pacman with extra features including an automatic architecture check that is only available in the base "cachyos" repository. The unmodified pacman in the Arch Linux "core" repository does not properly handle the CachyOS optimized repository architectures automatically and will likely run into issues installing packages from the optimized repositories.
> 
> To prevent any issues while manually migrating it is important to install the modified pacman from the "cachyos" repository as directed in Step 2 before adding the optimised repos to your pacman configuration and ensure the base "cachyos" repo is included when doing so.
> 
> Reconfiguring Arch's unmodified pacman to recognize the optimized repositories architectures manually is possible but not recommended or supported by the CachyOS team.
> :::

If this seems accurate to the devs i can submit a proper PR or y'all can just replace it yourselves if it's easier. I'd also suggest this warning should probably be moved under the "Manual" migration tab rather than the top of the whole migration section as it doesn't appear to be relevant when using the automatic migration script.

Aarrayy:
> If this seems accurate to the devs i can submit a proper PR or y'all can just replace it yourselves if it's easier. I'd also suggest this warning should probably be moved under the "Manual" migration tab rather than the top of the whole migration section as it doesn't appear to be relevant when using the automatic migration script.

I agree with you on this part. I moved it to the manual tab so it doesn't appear on the automatic tab.

gwuen:
> Or am i still completely misunderstanding something?

Indeed, I believe you are. There are incompatibilities both ways:

1. Upstream pacman will not install packages from the optimized repos because it doesn't recognize the architecture automatically. I wouldn't say that it “does not properly handle” custom architectures, it just requires manual intervention. You'd have to add them to the `Architecture` list in `/etc/pacman.conf`, but this *is not* mentioned on the wiki page.
2. Using CachyOS's pacman will add some custom entries to the local alpm database. When you revert to upstream pacman, this will lead to warnings because these custom entries aren't recognized. This *is* mentioned on the wiki page, along the advice to not install CachyOS's pacman if you want to avoid this.

That's also where my problem lies: Either paint the whole picture (i.e., explain how to use upstream pacman with the optimized repos) or remove this silly advice if you believe the optimized repos should only be used with CachyOS's pacman anyway.

Edit: Specifically, I'm in favor of removing this section that makes it sound like installing or not installing CachyOS's pacman are both supported options:

> If you want to avoid this, don't add the "cachyos" repository, which contains the customized pacman. All other repositories like cachyos-v3, cachyos-v4, cachyos-extra/core-v3/4 are safe to add.

PR #256, opened 2025-12-04
Add hardware acceleration flags for RTX 5070 TI with Brave

sky0matic:


Aarrayy:
Thanks for the addition!

PR #262, opened 2025-12-08
Only upgrade updated packages during optimized repo migration

pkwagner:
The current `pacman -Qqn | sudo pacman -S` command proposed for optimized repository migration re-installs all installed packages, no matter whether they have an optimized variant in one of the CachyOS repos or not (e.g., `core/hwdata`):

> warning: hwdata-0.401-1 is up to date -- reinstalling

Adding the `--needed` flag ignores same-version updates and only re-installs those that have a different version available:

> warning: hwdata-0.401-1 is up to date -- **skipping**


vnepogodin (approved):
Many thanks

PR #264, opened 2025-12-14
fix: update pacman version for optimized repos

azdanov:
See https://github.com/CachyOS/cachyos-repo-add-script/pull/8 also


vnepogodin (approved):
Many thanks

PR #269, opened 2025-12-20

natrius:
Added some information to the dual-boot

### DIFF
--- a/src/content/docs/installation/installation_on_root.mdx
+++ b/src/content/docs/installation/installation_on_root.mdx
@@ -298,7 +298,7 @@ One way to avoid this issue is to install each operating system on separate driv
 The EFI partition may need to be resized, depending on the recommended minimum size on the target boot manager. Refer to **[Manual Partitioning](/installation/installation_on_root#manual-partitioning)**. Note that FAT filesystems cannot be resized in-placed. To properly "resize" the FAT EFI partition, one must back up the existing EFI filesystem, use a tool such as `gparted` to resize the existing partition and reformat the new partition, then copy the previous filesystem structure back into the EFI partition.
 :::
 
-- Prerequisites
+### Prerequisites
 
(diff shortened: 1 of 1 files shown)


src/content/docs/installation/installation_on_root.mdx
```diff
@@ -318,6 +318,7 @@ The EFI partition may need to be resized, depending on the recommended minimum s
     - Press `Win + R`, type `diskmgmt.msc`, and press Enter to open Disk Management.
     - Identify your main Windows partition (usually C:) and right-click on it.
     - Click on `Shrink Volume...` and specify at least **30720MB** (30GB) and click on `Shrink`.
+- A +2GB EFI partition
```
1Naim, line 321: This is not the right place to put this. It's already written in manual partitioning and the previous sentence also links to it so I don't see why this should be here even if it makes sense.

src/content/docs/installation/installation_on_root.mdx
```diff
@@ -342,7 +343,7 @@ The EFI partition may need to be resized, depending on the recommended minimum s
             ├─nvme0n1p5 vfat       2G /boot
             └─nvme0n1p6 btrfs  239.6G /var/tmp
             ```
-            EFI partitions are almost always formatted as `FAT32/vfat`. Since `nvme0n1p1` does not have a Linux mount point, we can assume that this partition is the Windows EFI partition.
+            EFI partitions are almost always formatted as `FAT32/vfat`. Since `nvme0n1p1` does not have a Linux mount point, we can assume that this partition is the Windows EFI partition. Note: The EFI partition should be 2GB in size, otherwise the installation will fail.
```
1Naim, line 346: Again, doesn't make sense. The dual boot guide is performed after installation, so installation can't really fail after it has completed now can it?

PR #270, opened 2025-12-21
feat: update spanish

azdanov:


Aarrayy (approved):
Seems to be better than the previous batch.

PR #276, opened 2025-12-31

dwou:
Many changes: fixing language syntax and grammar, making text more terse, removing unnecessary text, reusing text where appropriate, fixing errors, other small changes.

Note: I did all this only editing plaintext, check syntax
Note: MBR/BIOS > GRUB doesn't have a boot partition step listed
Note: UEFI/GPT and MBR/BIOS can probably be consolidated
Note: "Double check that Install boot loader on: is pointing to /boot/efi" ~ where in the installation does it say "Install boot loader on:"? I can't find it.
Note: it sometimes uses 1000's of MiB and 1024's of MiB, should probably be consistent
Note: Removed the "In storage space constrained environments, ..." notes because the sizes were identical to the recommended sizes
Note: changed the "Installing multiple desktop environments ..." note from "danger" to "note" because there's no way to override this using calamares.

### DIFF
--- a/src/content/docs/installation/installation_on_root.mdx
+++ b/src/content/docs/installation/installation_on_root.mdx
@@ -12,26 +12,26 @@ import { Steps, Tabs, TabItem } from '@astrojs/starlight/components';
 
 ## Before you begin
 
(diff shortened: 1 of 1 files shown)


Aarrayy (approved):
Thanks a ton for the grammar fixes and corrections. My English improved a lot but clearly, there is room to improve. Happy new year!

PR #277, opened 2026-01-01

dwou:
Putting bold text in a title seems to break the formatting. See if this works.


Aarrayy (approved):
Yep, bold feature does not pair well in this case

PR #291, opened 2026-01-06
Gaming: Add entries about steam input for special controllers and Decky Loader

DavidRGriswold:
These were the only questions I still had about CachyOS Handheld Edition after installing it, and I've seen other people asking about both in the discord.

### DIFF
--- a/src/content/docs/configuration/gaming.mdx
+++ b/src/content/docs/configuration/gaming.mdx
@@ -465,6 +465,14 @@ Avoid using Proton on NTFS drives. Valve does not support this configuration, an
 If you're willing to proceed. Follow their **[unofficial guide](<https://github.com/ValveSoftware/Proton/wiki/Using-a-NTFS-disk-with-Linux-and-Windows>)**.
 :::
 
+### Enable full controller support on Steam for some controllers
+
+Some game controllers such as the 8Bitdo Ultimate 2 (in dinput mode) and the DualSense Edge require special rules to support features such as back buttons natively in Steam Input. If you have such a controller and are not getting full support, you can try copying the rules file `60-steam-input.rules` from [Valve's Github](https://www.github.com/ValveSoftware/steam-devices/) into the folder `/etc/udev/rules.d/` on CachyOS. This will require `sudo` or admin privileges.
+
+### [Decky Loader](https://decky.xyz)
+
+Decky Loader is used to add a variety of plugins and features to SteamOS machines. Though not officially supported on CachyOS it, and most plugins, generally works on CachyOS Handheld Edition, and may partially work in Steam Big Picture mode on a standard desktop as well.
+
 ## [Lutris](<https://lutris.net/>)
 
 Lutris is a game launcher on CachyOS. With Lutris, you can easily manage your game runners, including Wine, Proton, and emulators.


src/content/docs/configuration/gaming.mdx
```diff
@@ -465,6 +465,14 @@ Avoid using Proton on NTFS drives. Valve does not support this configuration, an
 If you're willing to proceed. Follow their **[unofficial guide](<https://github.com/ValveSoftware/Proton/wiki/Using-a-NTFS-disk-with-Linux-and-Windows>)**.
 :::
 
+### Enable full controller support on Steam for some controllers
+
+Some game controllers such as the 8Bitdo Ultimate 2 (in dinput mode) and the DualSense Edge require special rules to support features such as back buttons natively in Steam Input. If you have such a controller and are not getting full support, you can try copying the rules file `60-steam-input.rules` from [Valve's Github](https://www.github.com/ValveSoftware/steam-devices/) into the folder `/etc/udev/rules.d/` on CachyOS. This will require `sudo` or admin privileges.
```
1Naim, line 470: I'm not a fan of how this is written. Can you please rewrite it to be more generic, e.g.:

> If your controller doesn't work correctly via Steam Input, consider the following solutions:
> 1. Update the steam input udev rules from the [steam-devices repository](https://www.github.com/ValveSoftware/steam-devices/)
> <followed by a step-by-step or a quick command to download it and move it to /usr/lib/udev/rules.d>

The `/usr` path is preferred here compared to `/etc`, since the one in `/etc` quickly becomes an orphan once steam is updated, and also prioritized compared to the former, making controllers that are supported become unsupported because its missing rules.

Another solution which seems to also fix this specific case is to use the [game-devices-udev](https://aur.archlinux.org/packages/game-devices-udev) package. It complements the steam-devices ruleset really well and adds support for controllers that are not already in steam-devices.

src/content/docs/configuration/gaming.mdx
```diff
@@ -465,6 +465,14 @@ Avoid using Proton on NTFS drives. Valve does not support this configuration, an
...
+
+### [Decky Loader](https://decky.xyz)
+
+Decky Loader is used to add a variety of plugins and features to SteamOS machines. Though not officially supported on CachyOS it, and most plugins, generally works on CachyOS Handheld Edition, and may partially work in Steam Big Picture mode on a standard desktop as well.
```
1Naim, line 474: Again, not a fan of how this is written. There's not much info here other than "yes this software works on CachyOS". Either just remove this entirely, or expand it, e.g.:
a). add install instructions: instructions are already available in their own [website](https://wiki.deckbrew.xyz/en/user-guide/install), but we can add the quick install command here
b). useful plugins to use: I've heard DeckyPlumber be useful, and just recently a [plugin](https://github.com/jorgemmsilva/decky-plugin-fix-lego2-brightness-cachyos) to fix the brightness slider on the Legion Go 2

1Naim (changes requested):

src/content/docs/configuration/gaming.mdx
DavidRGriswold: I guess the reason I think this should be here is that I have seen it as an actual frequently asked question in discord - basically just "does it work?" Thus it feels like it might be worth putting in the Frequently Asked Questions.

But I can expand it with install instructions, sure.

src/content/docs/configuration/gaming.mdx
DavidRGriswold: Interesting, I didn't know I could use /usr instead of /etc, that's a better solution for sure. 

game-devices-udev did not solve the problem for my 8bitdo pro 2 controllers, but i could add a line that it may fix other controller problems.

1Naim: > game-devices-udev did not solve the problem for my 8bitdo pro 2 controllers

Strange, https://codeberg.org/fabiscafe/game-devices-udev/src/commit/a508fceff29e6f7ef5fac4478dc6285930b2040b/71-8bitdo-controllers.rules#L48-L57 it seems to be there.

PR #299, opened 2026-01-14
Add warning against using Pamac and Bauh

dwou:
Add warning to avoid Pamac and Bauh, and stick to Octopi/pacman

### DIFF
--- a/src/content/docs/configuration/post_install_setup.mdx
+++ b/src/content/docs/configuration/post_install_setup.mdx
@@ -10,6 +10,12 @@ import { Tabs, TabItem, Steps } from '@astrojs/starlight/components';
 
 ## Updating The System
 
+:::caution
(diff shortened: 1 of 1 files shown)


vnepogodin:
I think thats enough

https://wiki.cachyos.org/cachyos_basic/faq/#choosing-a-gui-package-manager

dwou:
In my opinion, if someone got to the bottom of that page then it's already too late.

Maybe a more general caution against improperly updating the system, and a pointer to Security & Best Practices, so more people are cognizant of package management and that Wiki section in particular, would be better?

dwou:
Something like that (I didn't test in a simulated environment)

Aarrayy:
Good idea, makes sense having a note or tip about that section in the (Updating System). I'll push a similar approach and i'll give you credits. Are you okay with that?

dwou:
Yes

Aarrayy:
There you go ae68e5a. Thanks!

PR #315, opened 2026-01-30
Use actual brackets instead of html codes

Numerlor:
Seems to have gotten mixed up in the last change that moved it into inline code from normal text. Currently displays the lt and gt literally on in the docs

https://wiki.cachyos.org/configuration/post_install_setup/#installing-and-using-the-cachyos-smbconf-file


Aarrayy:
You're right. Thanks

PR #316, opened 2026-01-31
Update recommended boot partition size to 4192

kylehendricks:
When using limine, the minimum is now 4192 MiB according to a new error that appears in the installer and the [26.01 changelog](https://wiki.cachyos.org/cachyos_basic/changelogs/gui_installer/#_top)


Aarrayy:
I don't see what's wrong, systemd-boot & rEFInd do not require 4192 MiB therefore both stayed at 2048, Limine has the correct requirement.

kylehendricks:
I'm an idiot. closing.

PR #321, opened 2026-02-13
i18n: add Romanian translation

BUSTheid:
I translated the whole wiki with Gemini and had a quick look over it - looks *okay* overall

Had to fix the internal links to point to the appropriate pages.

Will try to find the time to take an in-depth look at everything, but if there's any other Romanians who'd like to confirm that it's ready for merging - that would be much appreciated!

Still debating over Romanian-izing English technical terms vs full translation, and the usage of the dash in "-ul" vs simply adding the definite article onto untranslatable English terms, since it's subjective, so opinions about this appreciated as well.

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -111,6 +111,7 @@ export default defineConfig({
             ru: 'Начало работы',
             sk: 'Začíname',
             pt: 'Introdução',
+            ro: 'Primii Pași',
           },
           items: [
             {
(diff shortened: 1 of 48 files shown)


Aarrayy:
Thank you!, take your time

BUSTheid:
I pushed a commit where Gemini fixed the inconsistencies Copilot found in Romanian, but also another one in the other languages, where applicable easily, where the language didn't change the original English format.

BUSTheid:
Was very behind with upstream so I just nuked the branch and reinitialized it, ONLY focusing on the Romanian translation, without fiddling with inconsistencies/mistakes in other languages. I'll look over those again in a different PR, after this one is done.

PR #322, opened 2026-02-13
Change default ZSTD compression level in BTRFS documentation

Ririshi:
Update the default ZSTD compression level from 3 to 1 as it was recently changed in https://github.com/CachyOS/cachyos-calamares/pull/151


ptr1337:
This is only applied for NVME's

PR #324, opened 2026-02-14
Optimized Repositories: Updated dead link

TheHllm:
The previous link returns a 404. Which stops the command from succeeding.


TheHllm:
Nevermind, see #323

PR #325, opened 2026-02-15
Using sha256sum for one step verification of iso

mgeins:
Just using sha256sum -c instead of cat the checksum file and then verify. Just do it in one step with "sha256sum -c"


Aarrayy:
Neat suggestion. Thanks!

PR #331, opened 2026-02-17
Adjust the order of the sched-ext schedulers to match the sched-ext GUI

michaelr0:
Adjust the order of the sched-ext schedulers to match the sched-ext GUI


Aarrayy (approved):
Thanks!

PR #340, opened 2026-02-28
Remove outdated SDDM theme explanation due to sddm being switched out to plasmalogin by default

softwaresocialist:
Removed section explaining SDDM's default appearance.


Aarrayy:
Oh true. Thanks!

PR #345, opened 2026-03-08
Boot Manager Documentation Change

cameronbosch:
Updated the boot manager information for Handheld Edition, noting the change from systemd-boot to limine by default.


Aarrayy:
Thanks!

PR #357, opened 2026-03-16
Update FAQ with regular maintenance advice.

BombasticBooger:
Added section on CachyOS for regular maintenance advice.


Aarrayy (approved):
Good idea.

PR #369, opened 2026-03-25
Fixed typo in BTRFS Zswap instructions

NosYous:
This PR fixes a path inconsistency in the Btrfs swap file setup instructions.
In step 1 the guide creates a Btrfs subvolume at /swap. However, in step 4 for adding the entry to fstab, it incorrectly points to /swapfile (likely a carryover from the standard non-Btrfs guide when doing commit 9fb15bc).


NosYous:
Looks like it got fixed in a706735. Closing this PR.

PR #371, opened 2026-03-26
Update dual_gpu.mdx to include instructions to enable iGPU in Steam

jcnix:
Going through this documentation it was lacking some instruction on how to enable the iGPU for Steam. This can be desirable for greater battery life and reduced fan noise.

I only have access to a Radeon iGPU, so these instructions work for me. For an Intel iGPU the path to the Intel Vulkan driver would be substituted, but I don't have the ability to verify that.


Aarrayy:
Thanks!, if i manage to find something for Intel, i'll add it.

PR #373, opened 2026-03-29
Improve Windows hashsum check

lambdakilo:
AFAIK this should also support as far as Windows 7 PowerShell 2.0. Let me know if this is too narrow or too large scope for the possible new users of CachyOS.


lambdakilo:
Just checked it works on Windows 10 PowerShell 5.x too.

PR #374, opened 2026-03-31
socials: Remove Forum,Discord,Reddit

vnepogodin:
Due to demanding increase in moderation. Forum,Discord,Reddit are unable to handle moderation of these channels and therefore cannot be considered official where community rules must be enforced 

ps: also adds bluesky as official, which was missing previously

### DIFF
--- a/src/content/docs/support/social.md
+++ b/src/content/docs/support/social.md
@@ -1,6 +1,6 @@
 ---
 title: Official CachyOS Community
-description: Discord, Reddit, X, Forum, Mastodon
+description: Bluesky, Reddit, X, Forum, Mastodon
 ---
 
 The following Social Media are maintained by the CachyOS Team.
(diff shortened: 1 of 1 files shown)


ventureoo:
I just thought we could use discussions on GitHub in distribution repo / organization as a small forum for help or proposals. So we can also convert real user complaints into issues and mention them in PRs. GitHub is also much better protected against spam than our forum, which is constantly exposed to some attacks or conflicts, also no one from the team actively follows it. GitHub for obvious reasons is visited more frequently as it is directly related to distribution development, so I think there will be much less problems with moderating and it will allow our community to interact in a more cultural environment.

lambdakilo (changes requested):
Image

Reordering could be in place. I think this PR is good since you can only stretch so far as a maintainer.

I agree with @ventureoo that this GitHub Discussions could be an excellent idea since most of our code lives here anyway. At this popularity rate we probably need more communications maintainers. We could pin posts advertising that we need more maintainers to for example GitHub Discussions. I would also love to volunteer for such a role.

lambdakilo:
Also could be a good idea to find all instances where discord, reddit and forum/discourse are mentioned in all 115 repositories so that no traces are left there.

PR #383, opened 2026-04-13
Add ASUS motherboard caveats for secure boot setup

Narrox98:
This adds instructions and caveats for setting up and enabling Secure Boot on some ASUS motherboards, which behave differently from other vendors.

### Changes Made
Both the English and German versions of the wiki have been updated to include the following:

- Added a caution block explaining the risk of using `--firmware-builtin` on ASUS boards, as it can cause duplicate EFI key entries and lead to Secure Boot violations.
- Clarified that some ASUS boards lack a dedicated Secure Boot toggle and require `Windows UEFI Mode` to activate Secure Boot, while setting `OS Type` to `Other OS` actually disables it.
- Documented the need to use `--microsoft` instead and provided steps to re-enroll keys if necessary **(English wiki only)**.

### Context
- The changes are based on my observations from an ASUS Z790-P motherboard running a **dual-boot setup with Windows 11**.
- These instructions may not apply to all ASUS or other vendor boards, especially in non-dual-boot or different Linux-only setups.

### Question
**Why do the command instructions for enrolling keys differ between the English and German wiki versions?**

**German wiki:**
```bash
sudo sbctl enroll-keys --microsoft
```

**English wiki:**
 ```bash
 sudo sbctl enroll-keys --microsoft --firmware-builtin
 ```

### DIFF
--- a/src/content/docs/configuration/secure_boot_setup.mdx
+++ b/src/content/docs/configuration/secure_boot_setup.mdx
@@ -43,6 +43,10 @@ However, some MSI motherboards don't have a setup mode. To achieve the same effe
 <br />
 <ImageComponent imgsrc={import('~/assets/images/MSI-bios-secure-boot.jpg')} />
 
+Some Asus motherboards have a similar behavior to the above mentioned MSI motherboards, as they do not have a dedicated setup mode.
+
+Navigate to Boot → Secure Boot, set Secure Boot Mode to Custom, then open Key Management and select "Delete all Secure Boot Variables". 
+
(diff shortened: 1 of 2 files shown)


1Naim (approved):

1Naim:
> **Why do the command instructions for enrolling keys differ between the English and German wiki versions?**

German translation is not up-to-date with English.

vnepogodin:
image

Narrox98:
Yeah i missed the part of it being outdated my bad.

PR #386, opened 2026-04-15
docs: add notes about the outdated page warning

ptcodes:
While updating some outdated translated pages I spent a while trying to figure out where the "this page is outdated" banner was coming from.

Turns out it's Lunaria comparing the last git commit date of the English source file against the translation — if the source is newer, the banner appears automatically.

I've added a note to README.md explaining this so future contributors don't have to figure it out the hard way.


Aarrayy:
Good call.

PR #395, opened 2026-04-28
Remove UKUI from desktop environments section

divyansh-kataria:
UKUI is no longer available to install from the installer


Aarrayy:
You're correct. Thanks!

PR #411, opened 2026-05-07
Improvements in Gaming Section

lukaogadze:
THIS PR:
adds https://www.gamingonlinux.com/anticheat/ in game compatibility section
gaming on linux

adds "Fix videos not playing in Steam or game(s) displaying color bars"
unlock h264

adds log generation instruction for lutrus
lutris logs

@Aarrayy don't merge it without @loathingKernel 's approval

### DIFF
--- a/src/content/docs/configuration/gaming.mdx
+++ b/src/content/docs/configuration/gaming.mdx
@@ -400,6 +400,16 @@ They are not related and are irrelevant to each other. Our suggestion lately has
 - `Proton-GE` is a custom build made by [GloriousEggroll](https://github.com/GloriousEggroll/proton-ge-custom). It includes various fixes, and can be useful to have in certain situations.
 - `Proton 9.0.4 or lower` are the stable releases from `Valve`. Use this if the game you want to play only works with a previous Proton release.
 
+### Fix videos not playing in Steam or game(s) displaying color bars
+:::caution
+This is optional. Only use it if you are experiencing the issue.
+:::
+Start Steam with the following command. Wait for it to load, then exit Steam and start it normally.
+
+```sh
+ steam steam://unlockh264/
+```
+
 ### Fix Stuttering Caused by the Steam Game Recorder Feature
 
 :::caution
--- a/src/content/docs/configuration/gaming.mdx
+++ b/src/content/docs/configuration/gaming.mdx
@@ -379,7 +379,10 @@ CachyOS provides various Proton versions that are built and maintained for furth
 :::
 
 :::tip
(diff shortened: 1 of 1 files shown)


src/content/docs/configuration/gaming.mdx
```diff
@@ -397,6 +400,16 @@ They are not related and are irrelevant to each other. Our suggestion lately has
 - `Proton-GE` is a custom build made by [GloriousEggroll](https://github.com/GloriousEggroll/proton-ge-custom). It includes various fixes, and can be useful to have in certain situations.
 - `Proton 9.0.4 or lower` are the stable releases from `Valve`. Use this if the game you want to play only works with a previous Proton release.
 
+### Fix videos not playing in Steam or game(s) displaying color bars
```
loathingKernel: This is not a fix for all videos, this fixes h264 videos with Valve's Protons. AFAIK custom protons which provide media codecs do not need this. I think we should make the type of the media codec and that it affects official Protons clear somehow.

loathingKernel (changes requested):

src/content/docs/configuration/gaming.mdx
loathingKernel: `Fix H264 videos not playing or displaying color bars in Steam with Valve's Proton` would be my suggestion.

If I have assumed correctly so far, this setting won't affect any of the custom Protons, neither positively nor negatively

lukaogadze: @loathingKernel i did digging and did not managed to find "official" source for the 
```
steam steam://unlockh264/
```
it looks like it's some sort of leaked input parameter which can be found in github issues. it also fixed custom protons issues.

this guy got color bars in crimson desert when game was newly released. meanwhile other people using same proton did not.
color bars issue

https://youtu.be/rqU3s-YCSew?t=2050

i also talked with Linux-Next. no i haven't heard any issues occurring after enabling h264

color bars only appear in games 
maybe we should change it to:

```Fix H264 videos not playing or displaying color bars in Games with Valve's Proton```

loathingKernel: Proton-CachyOS will never show the color bars, it might display black if it can't decode the video, but will not display the color bars. The mediaconv part of proton is disabled completely by default. I suspect they used some other Proton, either by oversight or because they had set Proton-CachyOS as Steam's global defaul, and Steam used something else instead.

lukaogadze:
done @Aarrayy it's ready to merge
approved by proton god

loathingKernel (approved):

PR #422, opened 2026-05-20
docs(secure boot): Swap tab order Limine <> SystemD Boot in "Signing the Kernel Image and Boot Manager" section

ifeign:
Swapped `Signing the Kernel Image and Boot Manager` tab order to match other documents. As Limine is now the default boot loader in other places in the docs, it reads better to have the Limine tab selected by default for this page too. This matches the Destkop setup page in which the tab order is: Limine, SystemD/rEFInd, GRUB


Aarrayy:
Thanks!

PR #423, opened 2026-05-22
Update Limine configuration link to version 12.x

garthdw:
Updated the Limine configuration documentation to match the version of Limine now present on pacman for CachyOS.


Aarrayy:
Thanks for the update!

PR #493, opened 2026-06-16
Update post_install_setup.mdx

ettkatthej:
Improved the explanation of ufw allow / deny 
Added additional examples and text hinting at options to look up


Aarrayy:
Thanks, i agree too, they wasn't the most straightforward examples.

PR #515, opened 2026-07-05
noct v4-v5 migration guide

Jinra:
@azdanov I'm not sure how the formatting will look like with this. Are you able to preview to make sure it looks fine?


azdanov:
@Jinra I checked localy and fixed up a bit; can check after deploy if it's correct.

PR #530, opened 2026-07-21
Update hyprland.mdx

cantosun99:
While writing the German translation for this file I noticed a few issues and inconsistencies so I fixed a typo and a sentence structure and added periods where there should have been one to make the formating uniform.


Aarrayy:
Thanks!

PR #529, opened 2026-07-21
Create the German translation for hyprland.mdx

cantosun99:
Disclaimer: this is written by hand, not with an AI, but I don't use Hyprland myself so please verify if I interpreted something wrong!

PR #531, opened 2026-07-22
Change keybinds.lua to binds.lua

cantosun99:
As I wrote in the [comment on the German translation](https://github.com/CachyOS/wiki/pull/529#issuecomment-5042215493) in line 49 it still mentions the old fine keybinds.lua instead of the new binds.lua file. @Jinra is the fix I proposed correct?


ghost:
[Should be good to go](https://github.com/CachyOS/wiki/pull/529#issuecomment-5042923179)

PR #557, opened 2026-07-26
[DOCS]: Fix typo in Hyprland window management table in hyprland.mdx

sud33p:
Fixed typo where it says "Reize" instead of "Resize"


Aarrayy:
The typolice!

sud33p:
> The typolice!

More like typocd! 🥴

Issue #318, opened 2026-02-02
Edit warning about `game-performance` on old CPUs


dkchshv:
On Wiki written: "`game-performance` may not provide any benefit on old CPUs."
I think it should be something: "`game-performance` may not provide any benefit or do exactly the opposite on old CPUs."

Around 10-15% loss on 4C/8T

And in games `game-performance` affects performance even worse making game completely unplayable (by the way x86_64_v2 Proton does the same and makes CPU load up to 100% probably due to SSE4.X utilization).

ptr1337:
Well, CachyOS is mainly intended for modern hardware and if the Performance Profile provided by power-profiles-daemon then this is an issue with the Kernel Code from Intel maintaining the pstate driver

ghost:
Excuse me for bringing up this old issue, but do you perhaps have any kind of data as to where that cut-off roughly is? Either something like "around this or that generation" or "below this core-count/boost-frequency/single-core-/multi-core-performance"? I mean of course, in the end it's just a matter of trying it for yourself, but I'd love to know out of curiosity. And providing even a vague guideline in the wiki seems better than nothing!

Aarrayy:
I believe something in the lines of: older than 2017 CPU,  aka pre Ryzen and so on.

ghost:
That makes sense for AMD, what about Intel? Maybe specifying a cut-off point of 8th-gen (coffee-lake) and upwards since that's the point where they generally started to (mostly) ditch the quad-core CPUs?
Perhaps the wiki could specify something like this: "As with most options, you should test yourself if they improve performance on your specific system or not. As a rough guideline, anything below a 6-core CPU, so roughly around the AMD 1st-gen Ryzen and Intel 8th-gen (Coffee-Lake) CPUs can be seen as the cut-off."

PR #562, opened 2026-07-27
Edit warning about game-performance on old CPUs (Issue #318)

cantosun99:
This is a suggestion to implement what has been discussed in [Issue #318](https://github.com/CachyOS/wiki/issues/318). As the added text states, this is only an approximate guideline and the user is encouraged to try it out for himself.


Aarrayy:
Makes sense. ty

PR #576, opened 2026-08-03
switch_desktop.mdx: Updated Russian translation

cantosun99:
Also not much to say here, can't count that as a translation.


Aarrayy:
image

Issue #550, opened 2026-07-23
Keep translations outdated/missing or AI-translate them?


ghost:
As you've likely noticed when looking at the recent commits, I've been spending some time since yesterday to get the German translation from 24 done, 21 outdated, 1 missing to 38 done, 8 outdated, 0 missing. Now German is up to par with the Japanese translation as the currently two best translated languages but I plan to keep on going in the next few days to hopefully push it to completely up-to-date with English.

Compared to Japenese and German a lot of languages are starting to fall behind with some of the information being critically out-of-date to the point where I fear a non-English speaking user might come to the conclusion that CachyOS is a buggy mess that needs to be replaced with another distro.
Now my question is, should we vibe-translate the languages up to a point where the critical stuff is corrected or be patient and wait until someone decides to do it by hand like I'm doing right now with German?

I mean I do take pride in doing it 95% by hand, only using AI and then verifying afterwards when I don't really know the technical side of what I'm actually trying to translate, as was the case in a handful of sentences when translating the full hyprland file of which I have no clue about as a long time KDE user. However the question still stands, for the languages that don't have someone trying to keep them up-to-date with handwritten translations, is it better for them to stay outdated or to have a potentially inaccurate AI-translation?

If the consensus is that it should be done (I'm not just gonna pull request some ai-slop without anyone asking for it lmao) then the questions are which languages should be AI-translated and which ones should remain untouched, which particular files are critical enough to need to be translated and which ones aren't as important and how are they going to be translated. I personally could offer the time to do so and offer to use my Intel Arc Pro B70 with 32GB VRAM and 32GB DDR5 to run a local model for the translation, Qwen3.6 or Gemma 4 for example.

Aarrayy:
Being honest, the more we wait to translate, the worse it gets (duh) but at least we haven't added a lot of content in the past months but yeah, AI translation is a must since we barely have contributors, even myself that i'm a native Spanish speaker but i don't feel i would be able to do a "good enough" of a job translating.

SoongVilda:
@Aarrayy I would prefer the wiki to remain primarily English-only.

Most Linux commands, logs, warnings, and error messages are already in English, so users will generally need at least some English when troubleshooting. Maintaining multiple fully translated versions also creates a significant workload and makes it very easy for translations to become outdated or inconsistent with the English documentation.

AI-assisted translation could still be useful for making the wiki more accessible, especially for users who are not comfortable reading English. However, I think every translated version should include a clear notice such as:

[...]

PR #582, opened 2026-08-03
secure_boot_setup.mdx: Updated Russian translation

cantosun99:
As discussed in Issue #550 I will be starting to AI-translate the languages that don't have anyone translating them by hand like I'm doing with the German one. As most of the existing translations are done by AI anyway, this will just be an update, rather than a step back in quality.

I ran the wiki page local with bun and while I obviously can't verify the translation itself, I always try to verify if the site is visually identical to the English one and if things are or aren't translated.

---

I want the process to be as transparent as possible, which is why I'm listing prompt, model, soft- and hardware:

Prompt used: "Today we are going to translate some technical documentation. Specifically we are translating the wiki of the Arch-based Linux distribution CachyOS.

In general, the wiki is supposed to be English-first with the translations being there to complement it. A lot of the words and phrases should therefore be kept in English.
For example it makes no sense to translate things which are likely to be displayed in English anyway like an exemplary terminal output. Further, it makes no sense to translate the actual names of things into the requested language, "bootloader" being an example, as people know what a bootloader is.
Keeping that in mind, the overall goal isn't to literally translate every word, rather it's to translate the context around it so a native reader can grasp the concept and what he has to do. As an Arch-based Linux distro it's expected for the user to, at least on a very shallow level, to be aware of the concepts of a bootloader, a kernel and so on, but maybe the user doesn't have the reading-comprehension to understand what he has to do if he only reads the English wiki.

Structurally the wiki has the regular English files in a directory like /src/content/docs/configuration/gaming.mdx while the files in other languages have the exact same file name but are in a directory with an ISO language code like the corresponding German translation being in /src/content/docs/de/configuration/gaming.mdx. That has to be kept in mind when a link refers to another site of the wiki, as the language code has to be added in the translation, otherwise the link will send the user to the English page. For links that point outward of the wiki itself, they can simply be kept the same.

The translated pages usually already exist but are outdated by a few months compared to the main English ones, therefore a lot of the translation can and should be kept. You should apply all the differences: adding anything that was added, removing anything that has been removed and changing anything that was changed. However you should keep what wasn't changed so it's only an update with minimal lines changed, not a total rewrite of the entire page. The formating should be the exact same as the English one with every line-break and every punctuation mark being the exact same as that changes how the page is rendered. A ` stays a `, a ' stays a ' and so on, they are not interchangeable.

Now, we are going to translate the English file /home/c/Dokumente/GitHub/wiki/src/content/docs/configuration/secure_boot_setup.mdx to Russian which is the /home/c/Dokumente/GitHub/wiki/src/content/docs/ru/configuration/secure_boot_setup.mdx."

Model used: [A Qwen3.6 27B finetune](https://huggingface.co/DavidAU/Qwen3.6-27B-Fable-Fusion-711-Uncensored-Heretic-NM-DAU-NEO-MAX-MTP-GGUF/blob/main/Qwen3.6-27B-Fable-Fus-711-UnHeretic-NM-DAU-NEO-MAX-NEO-MTP-Q6_K.gguf) at Q6_K with Q8_0 KV-cache

Software used: My own AUR packages [llama.cpp-sycl](https://aur.archlinux.org/packages/llama.cpp-sycl) and [intel-deep-learning-essentials](https://aur.archlinux.org/packages/intel-deep-learning-essentials) to run the model, [Pi Coding Agent](https://pi.dev/) as the harness, running on CachyOS of course

Hardware used: Sparkle Intel Arc Pro B70 with 32GB VRAM


ghost:
/configuration is done, that concludes the Russian translations for today, I feel like the AI is doing a good job.

Time to continue working on my bachelors thesis after frying my GPU for two hours lmao

Aarrayy:
Your VRAM household is on fire. call the firefighters

ghost:
IMG_20260803_165152_617

The firefighters have already attached the water hoses!

PR #583, opened 2026-08-04
post_install_setup.mdx: Update disabling Cachy-Update

cantosun99:
I was browsing the CachyOS forum, saw [this post](https://discuss.cachyos.org/t/cachy-update-tray-icon/33716/2) and decided to update it really quick.

I also wanted to do the same to the German page in one PR, but I realized it's a complete mess and needs a whole rework anyway which I will get to eventually today.

### DIFF
--- a/src/content/docs/configuration/post_install_setup.mdx
+++ b/src/content/docs/configuration/post_install_setup.mdx
@@ -252,13 +252,12 @@ Includes a dynamic & clickeable systray applet for an easy integration with any 
 <details>
     <summary>How to disable Cachy-Update</summary>
         Open a terminal and run the following commands:
-            ```bash
-            sudo systemctl --global disable arch-update-tray.service
-            sudo systemctl --global disable arch-update.timer
+            ```sh
+            mkdir ~/.config/autostart
+            cp /etc/xdg/autostart/arch-update-tray.desktop ~/.config/autostart/
+            nano ~/.config/autostart/arch-update-tray.desktop
             ```
-            :::tip
-            If you want to keep the system tray applet enabled but just disable the automatic update check, you can simply disable the timer without disabling the service.
-            :::
+        Add `Hidden=true` at the end of the file, save, exit and reboot.
 </details>
 
 - Features:
--- a/src/content/docs/configuration/post_install_setup.mdx
+++ b/src/content/docs/configuration/post_install_setup.mdx
@@ -251,13 +251,16 @@ Includes a dynamic & clickeable systray applet for an easy integration with any 
 </details>
 <details>
     <summary>How to disable Cachy-Update</summary>
-        Open a terminal and run the following commands:
-            ```sh
-            mkdir ~/.config/autostart
-            cp /etc/xdg/autostart/arch-update-tray.desktop ~/.config/autostart/
-            nano ~/.config/autostart/arch-update-tray.desktop
-            ```
-        Add `Hidden=true` at the end of the file, save, exit and reboot.
+        Disable Cachy-Update in `CachyOS Hello > Apps/Tweaks > Cachy Update disabled`
+
+            If you just want to disable the systray applet, open a terminal and run the following commands:
+                ```bash
+                mkdir -p ~/.config/autostart
+                echo -e "[Desktop Entry]\nHidden=true" > ~/.config/autostart/arch-update-tray.desktop"
+                ```
+            :::tip
+            If you want to keep the system tray applet enabled but just disable the automatic update check, you can simply disable the timer without disabling the service.
+            :::
 </details>
 
 - Features:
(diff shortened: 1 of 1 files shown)


ghost:
There were two options discussed in the forum post, both make sense but I felt like this one looks more intuitive to the average user. What do you think which one should be included?

Antiz96:
Hi,

This could be simplified this way:

```
mkdir -p ~/.config/autostart
echo -e "[Desktop Entry]\nHidden=true" > ~/.config/autostart/arch-update-tray.desktop"
```

You want  `-p` with `mkdir` to avoid encountering an error in case the directory already exists.
There's no need to copy the `arch-update-tray.desktop` file beforehand. Simply creating one with the same name as an override containing just the "Hidden=true" parameter (and the mandatory [Desktop Entry] header of course) is sufficient.

Also, this desktop file only manages the systray applet. The instructions to disable the automated check timers still stands.

So:

- Do you wanna disable the automatic updates check?

```
sudo systemctl --global disable arch-update.timer
```

- Do you wanna disable the systray applet?

```
mkdir -p ~/.config/autostart
echo -e "[Desktop Entry]\nHidden=true" > ~/.config/autostart/arch-update-tray.desktop"
```

- Do you wanna disable both?

```
sudo systemctl --global disable arch-update.timer
mkdir -p ~/.config/autostart
echo -e "[Desktop Entry]\nHidden=true" > ~/.config/autostart/arch-update-tray.desktop"
```

I hope this helps :)

Antiz96:
Of course, another alternative is simply to remove the `cachy-update` package from the system (e.g. `sudo pacman -Rns cachy-update`). The above instructions are only relevant for people that wants to disable Cachy-Update's automated update checks and/or the systray applet but still want it to be installed anyway (to use it from the CLI for instance).

If you don't plan to use Cachy-Update at all, then you should just remove the package from the system.

ghost:
The suggested changes make a lot of sense, thank you! But your follow-up does as well -

> If you don't plan to use Cachy-Update at all, then you should just remove the package from the system.

Honestly, I thought the same when I wrote the commit. The program is described as "An update notifier & applier" so considering there are FOUR different options provided, you might as well choose another one. Also if you don't want the system tray notification you could just hide it here?

Bildschirmfoto_20260804_144231

Now the question is which direction to take for the change since the live version has to either be changed like you suggested or removed since it's kind of trivial to begin with.

Antiz96:
> Now the question is which direction to take for the change since the live version has to either be changed like you suggested or removed since it's kind of trivial to begin with.

I guess both directions could be represented maybe?

Something like :

> To uninstall Cachy-Update completely, do X
>
> If you wanna keep it installed but :
> - disable the systray applet, do Y
> - disable the automated checks, do Z

Or something like that.

ghost:
I applied the suggested changes and wow I feel dumb, you can just untick it in Cachy-Hello, just like the wiki tells you to do when you want to enable it, that will perform a clean uninstall of the four related packages.

To add more options, I've chosen to include your fix to just disable the applet and added the previous :::tip back in.

Now there are three up-to-date options to disable it.

Antiz96:
Indeed, unchecking it in from cachy-hello is the easiest way to remove it, nice catch!

Last details I would fix is that the tip for disabling the timer doesn't have the related command anymore and it still mention the service that doesn't exist anymore.

I would change it this way:

> If you want to keep the system tray applet enabled but just disable the automatic update check, disable the timer : sudo systemctl --global disable --now arch-update.timer

ghost:
I just wanted to check if that was still up-to-date while we are at it and noticed you are the maintainer of arch-update, that's fantastic haha!

In that case, is the whole "How to change the update check interval" section outdated? Seems on par with your documentation as far as I can tell.

Antiz96:
> I just wanted to check if that was still up-to-date while we are at it and noticed you are the maintainer of arch-update, that's fantastic haha!
> 

Hey hey ☺️ 

> In that case, is the whole "How to change the update check interval" section outdated? Seems on par with your documentation as far as I can tell.

Nearly! The method is still up-to-date but the default content of the file in the example is not anymore. I can share a suggestion of update later if you want (I'm AFK right now) :)

Antiz96:
Actually, you can just update the content of the file with the what's in here: https://github.com/CachyOS/cachy-update/blob/main/res/systemd/arch-update.timer

If you wanna keep the explanation comments: it check for updates 2 min after boot then once per day (with a 1 hour random delay). 

The rest is still correct :)

ghost:
aaalright, I think the third commit should cover everything then.

Thank you a lot for the help and for arch-update, I loved that package even on my Arch install before CachyOS!

Antiz96:
> aaalright, I think the third commit should cover everything then.
> 

Yup, looks good to me! Thanks a lot 🤗 

> Thank you a lot for the help and for arch-update, I loved that package even on my Arch install before CachyOS!

Thanks for the kind words, I appreciate it! 🙏

Issue #609, opened 2026-08-10
Translation banner's × button gets overlapped after scrolling on mobile


MintJapan:
https://github.com/user-attachments/assets/7b142e51-7d8c-45a6-ac53-438bb36a047a

As shown above, though it may seem like minor issue but this can be hassle on mobile.

Also, I think this should be translatable because leaving it as English may lead to confusion for non-English speakers.
(Yea I understand most Linux users are likely prepared for English, but thinking about beginners reading this wiki this is necessary I think)

@SoongVilda @cantosun99 What do you think about adding translations for it?

ghost:
1. Thank you for notifing me, I'll get to fixing this this evening or tomorrow morning! I set the banner to be fixed to the header but in your video the navigation bar does the same haha

2. I was debating if I should implement it in the specific language, but I feel like it should stay in English as people that can't really read English well won't benefit from the English wiki anyway. Still, I don't think it would confuse that kind of demographic as they will still likely have enough English knowledge to know what the banner roughly says and know to just dismiss it. I mean we are talking about an Arch-based distro, you WILL have to know just a little English.

So yes, I'll get to fixing 1 when I'm at my PC again, for point 2 idk, wouldn't even be hard to implememt it but I ultimately intentionally decided against it.

ghost:
If the consensus is that it should be then absolutely no problem, I'll get on it, but as I said, I personally would be in favor of keeping the banner English!

MintJapan:
> I feel like it should stay in English as people that can't really read English well won't benefit from the English wiki anyway.

I see your point, but if there's a major mistake in a translated article, readers who know that English is the authoritative source can run the original text through an AI to get the accurate details.
Also, it's a matter of consistency.

> Still, I don't think it would confuse that kind of demographic as they will still likely have enough English knowledge to know what the banner roughly says and know to just dismiss it. I mean we are talking about an Arch-based distro, you WILL have to know just a little English.

yeah I thought the same thing at first. However, considering how rapidly CachyOS has been growing in popularity nowadays, I think the wiki will be read by a much wider audience, going beyond the typical 'Arch distro' barrier.
It's still ranked the top on DistroWatch after all :D

ghost:
Hmm, good points honestly. I mean I already had a working version of that going so I'd say I get that one going again and let's wait what other people say?

ghost:
@MintJapan I already submitted the Fix for mobile in PR #610, did that on my laptop really quick!

[...]

PR #610, opened 2026-08-10
TranslationBanner.astro: Fixed Mobile Version

cantosun99:
Fix for Issue #609, the mobile banner now renders correctly.

The second part of the issue is still up for debate, any opinion would be appreciated!


ghost:
1200x675-1271987281

PR #611, opened 2026-08-10
TranslationBanner.astro: Added all languages

cantosun99:
Relevant to Issue #550 and #609 

@MintJapan Did my AI translate Japanese correct?

@Aarrayy @SoongVilda Do you think the banner should stay in English or support all the languages as in this PR?

Note: English is never actually used but kept as a fallback, Chinese is already added for the translation that I'm gonna do this week.

### DIFF
--- a/src/components/TranslationBanner.astro
+++ b/src/components/TranslationBanner.astro
@@ -8,18 +8,107 @@ if (!locale || isFallback) return null;
 const aiTranslated = entry.data.ai_translated ?? false;
 
 const hasOutdatedBanner = !!entry.data.banner;
+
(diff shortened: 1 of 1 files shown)


Aarrayy:
support all languages is ideal.

btw i noticed one thing in your spanish PR, the LLM duplicated an entire part which i noticed due to the automatic builder failing

check my last commit for reference.

ghost:
Haha, why on earth did it do that? Good catch, I manually scroll through the live English and new translated version side by side to identify anomalies visually and must have thought I didn't scroll far enough. I guess I'll verify lines of code first in the future!

Aarrayy:
sneaky LLM xD

PR #616, opened 2026-08-11
docs: add slow charging workaround for HP OmniBook X Flip

daviddelsol1998:
## What

Adds a "Workaround: Slow Charging on HP OmniBook X Flip" section to the Boot Manager Configuration page, covering the `acpi_osi=! acpi_osi="Windows 2022"` kernel command line workaround for Limine, GRUB, and systemd-boot.

## Why

The HP OmniBook X Flip 16-ar0xxx charges at ~0.2 W under Linux instead of 30 W+ because the EC firmware gates fast charging behind the ACPI `_OSI("Windows 2022")` query. CachyOS kernels don't provide `acpi_rev_override`, so the `_OSI` workaround is the way to fix it. Verified on a real device (charging restored to 30 W+).

* https://bugzilla.kernel.org/show_bug.cgi?id=221866
* https://github.com/CachyOS/linux-cachyos/issues/971

## Notes

* Uses the `KERNEL_CMDLINE[default]+=` append syntax (verified on device), plus GRUB and systemd-boot variants.
* Includes a verification section (`/proc/cmdline`, `BAT0/power_now`).
* Cautionary note that `acpi_osi=!` removes all Windows OSI responses.

### DIFF
--- a/src/content/docs/configuration/boot_manager_configuration.mdx
+++ b/src/content/docs/configuration/boot_manager_configuration.mdx
@@ -221,6 +221,78 @@ While entries are handled automatically, you can **configure the kernel paramete
 
    This command triggers the `mkinitcpio` process, which includes the `limine-mkinitcpio-hook`, ensuring your changes in `/etc/default/limine` are incorporated into the boot entries at `/boot/limine.conf`.
 
+## Workaround: Slow Charging on HP OmniBook X Flip
+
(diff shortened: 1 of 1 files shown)


Aarrayy:
Boot Manager Configuration page is not the correct one for this use case.

FAQ & Troubleshooting or in a new heading over at CachyOS Kernel.

PR #621, opened 2026-08-16
Fixed typo in Zram/Zswap documentation

Jinra:


ptr1337:
The change is not correct. It does disable zswap and not zram.

Jinra:
@ptr1337 ah , the verbiage of the wiki confused me and i though it said that creating the override file would _disable_ zswap. I understand now

PR #622, opened 2026-08-17
feat(i18n): add auto-redirect based on browser language

### LINKED ISSUE #416: [Feature Request] Switch language automatically
It would be better if it automatically redirect to translated pages based on user's browser language. 
On mobile, language switch button is located at the very bottom of the hamburger menu that difficult to find, so this will be especially helpful !

MintJapan:
**This change are made by Gemini 3.7 Flash.** Reviewed by me but I'm not well-experienced on web development, so it may contain mistakes or inappropriate implementations.

Close #416 .

---

## Summary
Automatically redirects users visiting English (root) pages to the translated version based on their browser language preferences (`navigator.languages`).

## Details
- **Zero content flash**: Injects a synchronous inline script into `<head>` to redirect before DOM paint.
- **Single source of truth**: Locales are derived from `lunaria.config.json` at build time.
- **Loop prevention & picker respect**: Uses `sessionStorage` and listens to language picker changes so manual selections aren't overwritten.
- **Clean history**: Uses `window.location.replace()` to keep browser history clean while preserving query parameters and hash fragments.
- **SEO safe**: Crawlers (Googlebot) and non-translatable paths (`/localization`, `/404`) are unaffected.

## Verification
- Build & lint checks pass (`bun run build`, `astro check`, `eslint`).
- Unit tested language detection & redirection flows.


ghost:
Hmm, I'm not sure if this goes against the notion in the banner that the English version is the authorative source. Instead of switching automatically, a simple "do you want to switch to the translated version?" suggestion as a notification would be more appropriate? The final choice is obviously to be made by the team, just my two cents on this!

MintJapan:
> Hmm, I'm not sure if this goes against the notion in the banner that the English version is the authoritative source.

@cantosun99 This implementation is designed to respect the user's choice during a session, so I don't think it will be much of an issue.
That said we could add a brief tip in the translation notice banner on how to switch languages (it's obvious on desktop, but on mobile it's tucked away inside the hamburger menu ≡). How about that?

ghost:
Oh my bad, I get it now, I read

> Automatically redirects users visiting English (root) pages to the translated version based on their browser language preferences

as it will do that even for users that visit the first time. If I now understand correctly that this simply saves their _concious decision_ to swap the language then I take back what I said!

MintJapan:
> Automatically redirects users visiting English (root) pages to the translated version based on their browser language preferences
> 
> as it will do that even for users that visit the first time. If I now understand correctly that this simply saves their _concious decision_ to swap the language then I take back what I said!

Oh sorry for the confusion, but your initial understanding was actually correct 🫥
It does automatically redirect users on their first visit based on browser language.

What I meant earlier by "designed to respect the user's choice" is that the language selector still works and users can switch back to English if needed, and it will be preserved throughout the session.

I don't want it to get in the way of reading the English wiki, but the current behavior feels pretty fine to me.

ghost:
I did have some time today to read the code and think about it and again, while I'm not part of the CachyOS team and they ultimately hold the right to decide, I think everyone should land on English by default and be forced to make the conscious decision to switch to another language. I appreciate the effort and the thought put into it, but I feel like users should be nudged to stay on English. I already tried to make the language select easier to find by moving it to the header of the hamburger menu on mobile instead of the footer so anyone should be able to switch if they really want to or need to.

MintJapan:
> I think everyone should land on English by default and be forced to make the conscious decision to switch to another language.

Rethinking this, I realized that the number of languages that are currently being properly maintained (by human) is actually quite limited.
In that situation, this feature may be inappropriate and could even become an unnecessary burden for users... 🫠

So, rather than adding the auto-redirect feature I’d like to change the approach and add a notice on English pages like you originally suggested.

Thank you for pointing out that issue!

ghost:
Yeah, if the translations were human-maintained like your Japanese one and my German one then I'd be all for it, but at the current state I just can't recommend using the translations, even the ones I'm doing with AI, because I'm just blindly trusting it to not be garbage. Again, certainly better than not having them at all, but we still shouldn't actively recommend them. Thank you for understanding!

PR #635, opened 2026-08-27
Unify GRUB EFI partition minimum size to 512 MiB

AtomicKaiser:
The Czech (cs) and Slovak (sk) installation_on_root guides specified a GRUB /boot/efi partition minimum of 100 MiB, while all other locales recommend 512 MiB. This inconsistency could lead to too-small EFI partitions when following the Czech/Slovak instructions.
This PR aligns the cs and sk manuals with the rest of the wiki by updating the GRUB EFI partition minimum to 512 MiB.

If you need any edits let me know.


AtomicKaiser:
I do not speak this tongues but all I did was change numbers so it should not impact it.

PR #643, opened 2026-09-15
Add note for account credentials in GPU migration guide

m1rza-s:
Added a note to make note of account credentials for TTY login during GPU migration.
Don't ask me why I think this is important :hidethepainharold:


Aarrayy:
You never know when a black screen will bite you.

ventureoo:
> You never know when a black screen will bite you.

I'm actually thinking about adding some protection from this in cachyos-drm-awaiter, so that you can at least get into tty automatically if GPU driver is failed.

